PR #13, opened 2020-01-09
STAC-7233 sap integration

TizianoPerrucci:


sap/tox.ini
```diff
@@ -0,0 +1,30 @@
+[tox]
+minversion = 2.0
+skip_missing_interpreters = true
+basepython = py27
+envlist =
+    {py27,py36}-sap
```
hruhek: Envlist `    {py27,py3}-sap` would loose Python 3 requirement, so it will work with versions 3.7 and 3.8.

TizianoPerrucci: nope, just tried, it will use the default python3 is on the path
if you want to test more version of py3 you need to be explicit and put `{py27,py34,py35,py36,py37}`

PR #15, opened 2020-01-24
Stac 6347 xray trace agent

hruhek:


aws/stackstate_checks/aws/aws.py
```diff
@@ -0,0 +1,290 @@
...
+            self.aws_secret_access_key = role['Credentials']['SecretAccessKey']
+            self.aws_session_token = role['Credentials']['SessionToken']
+
+    def get_account_id(self):
+        return self._get_boto3_client('sts').get_caller_identity().get('Account')
+
+    def get_xray_traces(self):
+        xray_client = self._get_boto3_client('xray')
+
+        start_time = datetime.datetime.utcnow() - datetime.timedelta(minutes=1)
```
zandrewitte: We might have to look at this a bit. Specifically this part: `datetime.timedelta(minutes=1)`. If the interval of the check is changed we might lose / duplicate data. For now this is probably ok, but can you make a follow up ticket to introduce the functionality to take the check execution time into account

zandrewitte: How hard would it be for you to implement this functionality? Otherwise we can also just do it now

hruhek: We can fix this that we record end time (prehaps in file) so in we can have something like:
`if old_end_time is None:`
`....old_end_time = datetime.datetime.utcnow() - datetime.timedelta(minutes=1)`
`start_time = old_end_time`
`end_time = datetime.datetime.utcnow()`

zandrewitte: Yup should work. Let's add it to the check configuration as a optional value and log when we update this `old_end_time` variable so if the checks fails you can restart it at that point

aws/README.md
```diff
@@ -0,0 +1,41 @@
+# Agent Check: Aws
+
+## Overview
+
+This check monitors [Aws][1] through the Datadog Agent.
```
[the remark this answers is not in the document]
hruhek: Fixed

aws/stackstate_checks/aws/aws.py
hruhek: Added caching old end time in file. If file is not defined in YAML config its created in temporary directory.

First iteration wrote json cache file, but differences in Python 2.7 and 3 made conversion of time format from string to datetime and back pain. So I used pickle, its in binary format, but no conversion or timezone voodoo is needed as it (de)serializes pure python objects.

hruhek: I implemented cache that is not in binary format as Tiziano requested.

PR #24, opened 2020-03-06
Less specific python dependency

whazenberg:
Binary python3.6 requirement relaxed to python3

I tried the commands on ZSH and ran into some minor issues, works fine in bash -- added that 'requirement'


TizianoPerrucci:
thank you

PR #31, opened 2020-03-23
if trigger disabled then send no event of problems

vishalxebia:


zabbix/stackstate_checks/zabbix/zabbix.py
```diff
@@ -288,25 +288,34 @@ def retrieve_problems(self, url, auth):
...
+            severity = item.get("severity", priority)
 
             zabbix_problem = ZabbixProblem(event_id, acknowledged, trigger_id, severity)
 
             self.log.debug("Parsed ZabbixProblem %s." % zabbix_problem)
             if not event_id or not trigger_id or not severity:
                 self.log.warn("Incomplete ZabbixProblem, got: %s" % zabbix_problem)
 
-            yield zabbix_problem
+            if priority is not None:
```
whazenberg: What is the case when there is no priority available? Add a comment and/or unit test?

zabbix/stackstate_checks/zabbix/zabbix.py
```diff
@@ -288,25 +288,34 @@ def retrieve_problems(self, url, auth):
...
+                yield zabbix_problem
 
     def get_trigger_priority(self, url, auth, trigger_id):
+        # `monitored` flag make sure we only get enabled triggers
```
whazenberg: Question; does that also include disabled triggers, the other support case?

zabbix/stackstate_checks/zabbix/zabbix.py
```diff
@@ -288,25 +288,34 @@ def retrieve_problems(self, url, auth):
...
         params = {
             "output": ["priority"],
-            "triggerids": [trigger_id]
+            "triggerids": [trigger_id],
+            "monitored": True
```
whazenberg: Unit test? for disabled and enabled triggers.

zabbix/stackstate_checks/zabbix/zabbix.py
vishalxebia: No, we will get only enabled triggers and this is what we want from the scenario.

zabbix/stackstate_checks/zabbix/zabbix.py
vishalxebia: if the trigger is enabled then for sure we will get the priority.

zabbix/stackstate_checks/zabbix/zabbix.py
vishalxebia: test case added.

zabbix/stackstate_checks/zabbix/zabbix.py
vishalxebia: put the comment for more clarification.

zabbix/tests/test_zabbix.py
```diff
@@ -493,3 +493,31 @@ def test_zabbix_respect_true_ssl_verify(self):
...
+        def _mocked_method_request(url, name, auth=None, params={}, request_id=1):
+            if name == "apiinfo.version":
+                return self._apiinfo_response()
+            elif name == "host.get":
+                return self._zabbix_host_response()
+            elif name == "problem.get":
+                response = self._zabbix_problem()
+                return response
+            elif name == "trigger.get":
+                return None
```
whazenberg: Does this work properly? Given that the code is:

```
response = self.method_request(url, "trigger.get", auth=auth, params=params)
trigger = response.get('result')
```

I'm expecting the response to be `response['result'] = []`

vishalxebia: @whazenberg  correct I was wrong in this. I have made the change and also there should be an event for a host without any problems/triggers which will make the host healthy.

PR #38, opened 2020-04-16
static topology check migration from v1

vishalxebia:


static_topology/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Configuration
+
+1. Edit the `static_topology.d/conf.yaml` file, in the `conf.d/` folder at the root of your
+   Agent's configuration directory to start collecting your static_topology performance data.
+   See the [sample static_topology.d/conf.yaml][1] for all available configuration options.
+
+2. Restart the Agent
+
+## Data Collected
```
TizianoPerrucci, line 22: Can you update this section with relevant information ?

static_topology/stackstate_checks/static_topology/static_topology.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def get_instance_key(self, instance):
+        if "components_file" not in instance:
+            raise ConfigurationError('Static topology instance missing "components_file" value.')
+
+        return TopologyInstance(self.INSTANCE_TYPE, instance["components_file"])
+
+    def check(self, instance):
+        if 'components_file' not in instance:
+            raise ConfigurationError('Static topology instance missing "components_file" value.')
```
TizianoPerrucci, line 23: we don't throw exception from the check but rather use the `service_check` critical

static_topology/stackstate_checks/static_topology/static_topology.py
```diff
@@ -0,0 +1,136 @@
...
+        if 'relations_file' not in instance:
+            raise ConfigurationError('Static topology instance missing "relations_file" value.')
```
TizianoPerrucci, line 25: we don't throw exception from the check but rather use the `service_check` critical

static_topology/stackstate_checks/static_topology/static_topology.py
```diff
@@ -0,0 +1,136 @@
...
+        if 'type' not in instance:
+            raise ConfigurationError('Static topology instance missing "type" value.')
```
TizianoPerrucci, line 27: we don't throw exception from the check but rather use the `service_check` critical

static_topology/stackstate_checks/static_topology/static_topology.py
```diff
@@ -0,0 +1,136 @@
...
+            except ConfigurationError as e:
+                self.log.error(str(e))
+                raise e
+            except Exception as e:
+                self.log.exception(str(e))
+                self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL, message=str(e), tags=instance_tags)
+            finally:
+                self.stop_snapshot()
+        else:
+            raise ConfigurationError('Static topology instance only supports type CSV.')
```
TizianoPerrucci, line 50: we don't throw exception from the check but rather use the `service_check` critical

static_topology/stackstate_checks/static_topology/static_topology.py
```diff
@@ -0,0 +1,136 @@
...
+
+                # identifiers processing
+                identifiers = data.get('identifiers', "")
+                # identifiers column may be in the row but may be empty/unspecified for that row, defaulting
+                identifiers = identifiers.split(',') if identifiers else []
+                data['identifiers'] = identifiers
+
+                self.component(row[id_idx], row[type_idx], data)
+
+    def handle_relation_csv(self, filelocation, delimiter, instance_tags):
```
TizianoPerrucci, line 106: seems we don't use `instance_tags` here, should we?

static_topology/stackstate_checks/static_topology/static_topology.py
vishalxebia: @TizianoPerrucci  this is the config param check and we always raise `ConfigurationError`.

static_topology/stackstate_checks/static_topology/static_topology.py
vishalxebia: @TizianoPerrucci We don't create `service_checks` for config params missing. As an example:- https://github.com/StackVista/stackstate-agent-integrations/blob/04230dabb5062f23a643a46d644d71172c5479b4/sap/stackstate_checks/sap/sap.py#L37

static_topology/stackstate_checks/static_topology/static_topology.py
vishalxebia: @TizianoPerrucci We don't create service_checks for config params missing.

static_topology/stackstate_checks/static_topology/static_topology.py
vishalxebia: this is removed and converted into `service_checks`.

PR #40, opened 2020-04-28
servicenow migration

vishalxebia:


servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -0,0 +1,199 @@
...
+        instance_tags = instance_config.instance_tags
+
+        state = self._collect_components(instance_config, timeout)
+
+        for component in state['result']:
+            comp_name = component['name'].encode('utf-8')
+            external_id = str(comp_name) if isinstance(comp_name, bytes) else comp_name
+            comp_type = component['sys_class_name']
+            data = {
+                "sys_id": component['sys_id'],
```
zandrewitte, line 102: comp_name is only used in the externalId? What is used for the name? Can we also generate a urn based identifer for the components here? Something like: 
`urn:service-now:type:identifier` 
Is it possible that things from ServiceNow should merge with other StackPacks?

vishalxebia: @zandrewitte I created the identifier on basis of `sys_id` and format is `urn:servicenow:type:sys_id`..

PR #41, opened 2020-05-04
Stac 8556 nagios integration

hruhek:


nagios/stackstate_checks/nagios/data/conf.yaml.example
```diff
@@ -0,0 +1,9 @@
+init_config:
+  min_collection_interval: 30
```
TizianoPerrucci, line 2: in agent v2 this config is part of instance https://docs.datadoghq.com/developers/write_agent_check/?tab=agentv6v7#collection-interval

I would remove it

nagios/stackstate_checks/nagios/nagios.py
```diff
@@ -0,0 +1,442 @@
...
+                tailers = []
+                nagios_conf = {}
+                instance_key = None
+
+                if 'nagios_conf' in instance:  # conf.d check
+                    conf_path = instance['nagios_conf']
+                    nagios_conf = self.parse_nagios_config(conf_path)
+                    instance_key = conf_path
+                # Retrocompatibility Code
+                elif 'nagios_perf_cfg' in instance:
```
TizianoPerrucci, line 90: this is not mentioned in the `conf.example.yaml`, would be good to describe it

nagios/stackstate_checks/nagios/nagios.py
```diff
@@ -0,0 +1,442 @@
...
+                    conf_path = instance['nagios_perf_cfg']
+                    nagios_conf = self.parse_nagios_config(conf_path)
+                    instance["collect_host_performance_data"] = True
+                    instance["collect_service_performance_data"] = True
+                    instance_key = conf_path
+                if 'nagios_log' in instance:
```
TizianoPerrucci, line 96: this is not mentioned in the `conf.example.yaml`, would be good to describe it

nagios/stackstate_checks/nagios/nagios.py
```diff
@@ -0,0 +1,442 @@
...
+                    tailers.append(NagiosEventLogTailer(
+                        log_path=nagios_conf['log_file'],
+                        file_template=None,
+                        logger=self.log,
+                        hostname=self.hostname,
+                        event_func=self.event,
+                        gauge_func=self.gauge,
+                        freq=check_freq,
+                        passive_checks=instance.get('passive_checks_events', False)))
+                if 'host_perfdata_file' in nagios_conf and \
```
TizianoPerrucci, line 116: this is not mentioned in the conf.example.yaml, would be good to describe it

nagios/stackstate_checks/nagios/nagios.py
```diff
@@ -0,0 +1,442 @@
...
+                    self.log.debug("Starting to tail the host_perfdata file")
+                    tailers.append(NagiosHostPerfDataTailer(
+                        log_path=nagios_conf['host_perfdata_file'],
+                        file_template=nagios_conf['host_perfdata_file_template'],
+                        logger=self.log,
+                        hostname=self.hostname,
+                        event_func=self.event,
+                        gauge_func=self.gauge,
+                        freq=check_freq))
+                if 'service_perfdata_file' in nagios_conf and \
```
TizianoPerrucci, line 128: this is not mentioned in the conf.example.yaml, would be good to describe it

nagios/tests/test_e2e.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+
+@pytest.mark.e2e
+def test_e2e(sts_agent_check, instance):
+    # This integration is based on tailing logs but because there is an agent running
```
TizianoPerrucci, line 6: I dont' understand this comment, can you rephrase it ?

hruhek: This is file copied from DataDog repo.
https://github.com/DataDog/integrations-core/blob/master/nagios/tests/test_e2e.py

hruhek: Datadog had problem with running tests because of Tailer missing its offset when reading log files. 
We encountered the same problem. 

Tailing logs works only when agent is running and starting check. If you invoke check manually Tailer will miss the offset and won't be able to parse logs.

Would this remark clarify it?

nagios/stackstate_checks/nagios/data/conf.yaml.example
hruhek: Removed

nagios/stackstate_checks/nagios/nagios.py
hruhek: I searched in Nagios docs but could not find this `nagios_perf_cfg` setting. 
I found it only in DataDog Nagios check and here https://github.com/openstack/monasca-agent/blob/master/packaging/monasca-agent/win32/install_files/datadog_win32.conf

I would guess it's from Nagios 3 configuration.
We didn't test check with that Nagios version. Its last release was in 2013.
https://www.nagios.org/projects/nagios-core/history/3x/

Should we remove that Retrocompatibility Code?

TizianoPerrucci: lets keep notes about supporting old version

nagios/stackstate_checks/nagios/nagios.py
TizianoPerrucci: lets keep notes about supporting old version

nagios/tests/test_e2e.py
TizianoPerrucci: The new message is indeed clearer

nagios/stackstate_checks/nagios/nagios.py
hruhek: I added comment with retro compatibility options to `conf.yaml.example`

nagios/tests/test_e2e.py
hruhek: As agreed yesterday `test_e2e.py` removed.

PR #43, opened 2020-06-16
catch the exception for metadata not present and continue

vishalxebia:


vsphere/stackstate_checks/vsphere/vsphere.py
```diff
@@ -989,22 +989,24 @@ def _vsphere_vms(self, content, domain="Unspecified", regexes=None, include_only
...
+                    if isinstance(c, vim.VirtualMachine):
+                        topology_tags["topo_type"] = VSPHERE_COMPONENT_TYPE.VM
+                        topology_tags["name"] = c.name
+                        topology_tags["layer"] = TOPOLOGY_LAYERS.VM
+                        topology_tags["domain"] = domain
+
+                        sts_identifiers, labels = self.extract_tags("VirtualMachine", c._moId)
+                        topology_tags["identifiers"] = sts_identifiers
+
+                        topology_tags["datastore"] = c.datastore[0]._moId
```
mpvvliet, line 1002: What happens if `c.datastore` is empty?

vsphere/stackstate_checks/vsphere/vsphere.py
```diff
@@ -1021,35 +1023,37 @@ def _vsphere_datacenters(self, content, domain="Unspecified", regexes=None, incl
...
-                topology_tags["clustercomputeresources"] = clustercomputeresources
-
-                add_label_pair(labels, "name", topology_tags["name"])
-                hostname = None
+                    computeresources = []
+                    clustercomputeresources = []
+                    datastores = []
+                    for datastore in c.datastore:
+                        datastores.append(datastore.name)
+                    topology_tags["topo_type"] = VSPHERE_COMPONENT_TYPE.DATACENTER
```
mpvvliet, line 1041: I would set the type above with the other required parameters (name, id, layer, etc.). Otherwise, if an error occurs in the code between setting the identifiers and the type, we could end up with a component that misses a type.

vsphere/stackstate_checks/vsphere/vsphere.py
vishalxebia: @mpvvliet it will go in exception and continue the check with creating the component and there is no datastore associated with it.

vsphere/stackstate_checks/vsphere/vsphere.py
```diff
@@ -989,22 +989,24 @@ def _vsphere_vms(self, content, domain="Unspecified", regexes=None, include_only
...
+                        topology_tags["datastore"] = c.datastore[0]._moId
+                        add_label_pair(labels, "name", topology_tags["name"])
+                        add_label_pair(labels, "guestId", c.config.guestId)
+                        add_label_pair(labels, "guestFullName", c.config.guestFullName)
+                        add_label_pair(labels, "numCPU", c.config.hardware.numCPU)
+                        add_label_pair(labels, "memoryMB", c.config.hardware.memoryMB)
+                except Exception as e:
+                    self.log.exception("Exception occurred during collecting metadata for VMs: {}".format(str(e)))
                 topology_tags["labels"] = labels
                 obj_list.append(dict(mor_type="vm", hostname=hostname, topo_tags=topology_tags))
```
mpvvliet, line 1011: If I read the code correctly, we are still sending a component even if an exception occurs. Depending on where the exception happened (either when we were still determining the required parameters (id, type, name, etc) or after that) we may or may not have a complete component definition.

I suggest you separate the retrieval of mandatory fields (and handle the exception by not sending a component) from optional fields (and handle the exception by sending the component and metadata we have collected).

vsphere/stackstate_checks/vsphere/vsphere.py
```diff
@@ -1066,29 +1070,31 @@ def _vsphere_datastores(self, content, domain="Unspecified", regexes=None, inclu
...
-                topology_tags["accessible"] = c.summary.accessible
-                topology_tags["capacity"] = str(c.summary.capacity)
-                topology_tags["type"] = c.summary.type
-                topology_tags["url"] = c.summary.url
-                topology_tags["layer"] = TOPOLOGY_LAYERS.DATASTORE
-                topology_tags["domain"] = domain
-
-                sts_identifiers, labels = self.extract_tags("Datastore", c._moId)
-                topology_tags["identifiers"] = sts_identifiers
+                    sts_identifiers, labels = self.extract_tags("Datastore", c._moId)
```
mpvvliet, line 1084: Suggest you group the required attributes first and collect others (starting from `accessible`) later so that we have a complete component even if fetching `accessible` or `capacity` fails.

vsphere/stackstate_checks/vsphere/vsphere.py
mpvvliet: Have you considered rewriting it using a `for` loop rather than straight indexing element `0`? Then if it's empty it will at least continue and we will get further information as well.

PR #46, opened 2020-07-03
stac-9512: support filter resource type for servicenow

vishalxebia:


servicenow/tests/test_servicenow.py
```diff
@@ -210,3 +199,83 @@ def test_get_json(self, mock_req_get):
...
+        self.check.get_sys_class_component_filter_query.return_value = "sysparm_query=sys_class_namecmdb_ci_netgear" \
+                                                                       "%2Ccmdb_ci_cluster%2Ccmdb_ci_app_server"
+        self.check._get_json = mock.MagicMock()
+        self.check._get_json.return_value = json.loads(mock_collect_components())
+        self.check._process_components(instance_config, 10)
+
+        topo_instances = topology.get_snapshot(self.check.check_id)
+        self.assertEqual(len(topo_instances['components']), 1)
+        self.assertEqual(len(topo_instances['relations']), 0)
+        self.assertEqual(topo_instances['components'][0]['type'], 'cmdb_ci_computer')
```
zandrewitte, line 250: Not sure why this is returned when it looks like the query should only return: 
```
"cmdb_ci_netgear", "cmdb_ci_cluster", "cmdb_ci_app_server"
```

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -138,13 +187,17 @@ def _collect_component_relations(self, instance_config, timeout, offset, batch_s
...
+        sys_class_filter_query = self.get_sys_class_relation_filter_query(sys_class_filter)
+        if sys_class_filter_query:
+            url = url + "?{}".format(sys_class_filter_query)
+            self.log.debug("URL for relation collection after applying filter:- %s", url)
 
         return self._get_json_batch(url, offset, batch_size, timeout, auth)
 
     def _process_component_relations(self, instance_config, batch_size, timeout, relation_types):
         offset = 0
-        instance_tags = instance_config.instance_tags
```
zandrewitte, line 147: Why did this get dropped?

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -95,14 +138,20 @@ def _process_components(self, instance_config, timeout):
         state = self._collect_components(instance_config, timeout)
 
         for component in state['result']:
-            comp_name = component['name'].encode('utf-8')
+            component = self.filter_empty_metadata(component)
+            identifiers = []
+            comp_name = component['name']
             comp_type = component['sys_class_name']
-            external_id = "urn:servicenow:{}:{}".format(comp_type, component['sys_id'])
```
zandrewitte, line 100: why did this change into `component['sys_id']`

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -95,14 +138,20 @@ def _process_components(self, instance_config, timeout):
...
-            data = {
-                "sys_id": component['sys_id'],
```
zandrewitte, line 102: is the `sys_id` now always in the returned data?

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: Yeah because we get only sys_id in the relation API response and need to create relation.

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: yes it's a mandatory attribute which will be always present for every component.

PR #49, opened 2020-07-29
Use only SAPHostAgent and fixed cert authentication.

DennisLoos:
No Jira issue

I rewrote the existing checks to only use SAPHostControl. This simplifies connection to the SAP system.

Also fixed certificate authentication so the basic authentication credentials are not sent as well.

- [ ] Unit/Component test
- [ ] Integration test	

I did not change any tests.

I got a PO / QA Approval by:
- [ ] Name

I tested it in our K8S production cluster as part of the onboarding of the first customer.

I am already using it myself in our cluster as a work around.


TizianoPerrucci:
@DennisLoos thanks for the PR

Unfortunately we just figured that the pipeline is not working for PR created by forks (is not your problem, you are doing it right, we need to somehow fix it on our end)

Never the less I run locally the `sap` check tests and they are failing:
```
$ checksdev test --cov sap
...
tests/test_sap.py::test_collect_processes FAILED
tests/test_sap.py::test_collect_worker_metrics FAILED
tests/test_sap.py::test_collect_memory_metric FAILED
```

is not strange they are failing because you are not using the `SAPHostAgent` anymore so the processes and metrics tests should mock a different response.

Let us know if you need any help with it

DennisLoos:
Worked on Tiziano's comments and commited new code. Will make a new PR for the latest code.

PR #57, opened 2020-09-08
stac 10422: add domain, environment and tags coming from config

vishalxebia:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


sap/stackstate_checks/sap/sap.py
```diff
@@ -57,6 +59,8 @@ def _get_config(self, instance):
         self.verify = instance.get("verify", True)
         self.cert = instance.get("cert", "")
         self.keyfile = instance.get("keyfile", "")
+        self.domain = instance.get("domain", "")
+        self.stackstate_environment = instance.get("environment", "")
```
zandrewitte, line 63: @vishalxebia If we don't set the instance domain and environment it will default to empty strings instead of None. Can we keep it None instead such that we can decide whether we want to send it to StackState or not and have the StackPack put the default value. Or put the default value in here instead of an empty string

sap/tests/conftest.py
```diff
@@ -21,6 +21,8 @@ def instance():
         "url": "http://localhost",
         "user": "test",
         "pass": "test",
+        "domain": "sap",
+        "environment": "sap-prod"
     }
```
zandrewitte, line 26: Missing the configuration for `tags`

sap/tests/test_sap.py
```diff
@@ -34,7 +34,8 @@ def test_cannot_connect_to_host_control(aggregator, instance):
             instance_key=TopologyInstance("sap", "LAB-SAP-001"),
             components=[
                 {"id": "urn:host:/LAB-SAP-001", "type": "sap-host",
-                 "data": {"host": "LAB-SAP-001", 'tags': ['integration-type:sap', 'integration-url:LAB-SAP-001']}}
+                 "data": {"host": "LAB-SAP-001", 'tags': ['integration-type:sap', 'integration-url:LAB-SAP-001'],
```
zandrewitte, line 37: Assert that instance tags get added here and the following assertions

zandrewitte (changes requested):

sap/stackstate_checks/sap/sap.py
vishalxebia: @zandrewitte I had put `None` earlier but not sure what it is translated into JSON response for the replay. So I reverted and put back to an empty string. For an empty string, I am sure it works properly.

zandrewitte (approved):

PR #61, opened 2020-10-09
Stac 10855 servicenow bug

vishalxebia:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/servicenow/stackstate_checks/servicenow/servicenow.py
+++ b/servicenow/stackstate_checks/servicenow/servicenow.py
@@ -62,11 +62,10 @@ class ServicenowCheck(AgentCheck):
             relation_types = self._process_and_cache_relation_types(instance_config, timeout)
             self._process_components(instance_config, timeout)
             self._process_component_relations(instance_config, batch_size, timeout, relation_types)
+            self.stop_snapshot()
         except Exception as e:
             self.log.exception(str(e))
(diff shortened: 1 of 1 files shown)


servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -137,23 +142,26 @@ def _process_components(self, instance_config, timeout):
...
-                identifiers.append("urn:host:/{}".format(component['fqdn']))
-            if 'host_name' in component and component['host_name']:
-                identifiers.append("urn:host:/{}".format(component['host_name']))
-            identifiers.append("urn:host:/{}".format(comp_name))
-            identifiers.append(external_id)
-            data = {"identifiers": identifiers, "tags": instance_tags}
-            data.update(component)
-
-            self.component(external_id, comp_type, data)
+        if "result" in state:
```
zandrewitte, line 145: I think you also have to check the contents of `result`. In the error log it looked like it ended up being a string, is that still possible now?

zandrewitte (approved):

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: I am already checking the content of result is `dictionary` or not.

PR #66, opened 2020-10-16
powershell alerts feature

melnemr:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/scom/stackstate_checks/scom/scom.py
+++ b/scom/stackstate_checks/scom/scom.py
@@ -3,8 +3,9 @@ import base64
 from requests_ntlm import HttpNtlmAuth
 import time
 import re
+import json
 from stackstate_checks.base import AgentCheck, ConfigurationError, TopologyInstance
-
+import subprocess
 try:
     from urlparse import urlparse
 except ImportError:
@@ -55,7 +56,24 @@ class SCOM(AgentCheck):
 
         instance_url = instance['hostip']
         return TopologyInstance(self.INSTANCE_TYPE, instance_url)
-
+    def sendAlertsPowershell(self,scom_server):
+        cmd = subprocess.Popen(["powershell.exe","Start-OperationsManagerClientShell -managementServerName "+scom_server+"\n","Get-SCOMAlert | select \"monitoringobjectid\", \"name\", \"monitoringobjectdisplayname\", \"description\", \"resolutionstate\", \"monitoringobjectpath\" -ExpandProperty  timeadded| ConvertTo-Json \n"],stdout=subprocess.PIPE)
+        out = cmd.communicate()[0]
+        alerts = json.loads(out)
+        for event in alerts:
+            self.event({
+                "timestamp": int(time.time()),
+                "msg_title": event.get("Name"),
+                "msg_text": event.get("Description"),
+                "source_type_name": "Alert",
+                "host": event.get("MonitoringObjectDisplayName"),
+                "tags": [
+                    "id:%s" % event.get("MonitoringObjectId"),
+                    "server:%s" % event.get("MonitoringObjectPath"),
+                    "resolution_state:%s" % event.get("ResolutionState"),
+                    "time_added:%s" % event.get("DateTime")
+                ]
+            })
     def sendAlerts(self, session, scom_ip, domain, username, password, component_id):
         data = {
             "criteria": "(MonitoringObjectId = '" + component_id + "')",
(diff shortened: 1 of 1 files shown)


scom/stackstate_checks/scom/scom.py
```diff
@@ -55,7 +56,24 @@ def get_instance_key(self, instance):
 
         instance_url = instance['hostip']
         return TopologyInstance(self.INSTANCE_TYPE, instance_url)
-
+    def sendAlertsPowershell(self,scom_server):
+        cmd = subprocess.Popen(["powershell.exe","Start-OperationsManagerClientShell -managementServerName "+scom_server+"\n","Get-SCOMAlert | select \"monitoringobjectid\", \"name\", \"monitoringobjectdisplayname\", \"description\", \"resolutionstate\", \"monitoringobjectpath\" -ExpandProperty  timeadded| ConvertTo-Json \n"],stdout=subprocess.PIPE)
+        out = cmd.communicate()[0]
+        alerts = json.loads(out)
```
lowdewijk, line 62: In other checks `yaml.loads` is used. You should also be aware of what Unicode encoding is used. Any idea?

scom/stackstate_checks/scom/scom.py
```diff
@@ -92,23 +110,30 @@ def getComponentDetails(self, id, session, scom_ip, domain, username, password):
...
         ids = []
         response = (session.get('http://' + scom_ip + '/OperationsManager/data/objectInformation/' + serviceid,
                                 auth=HttpNtlmAuth(domain + '\\' + username, password))).json()
         data = response.get("relatedObjects", [])
         for idd in data:
             ids.append(idd.get("id"))
+            type_data = {
+                            "criteria": "(Id = '"+str(idd.get("id"))+"')"
+                            }
+            type_response = (session.post('http://'+ scom_ip +'/OperationsManager/data/scomObjects',
```
lowdewijk, line 123: It is much better to use the `base_url` that includes the protocol. You should then create a function to creates a valid SCOM URL.. without duplication of slashes!

scom/stackstate_checks/scom/scom.py
```diff
@@ -122,16 +147,17 @@ def nextNodes(self, serviceid, session, scom_ip, domain, username, password, sys
                 ]
             })
             self.relation(str(serviceid), str(idd.get("id")), "hosted_on", {})
-            self.sendAlerts(session, scom_ip, domain, username, password, idd.get("id"))
+            if "api" in integration_mode:
+               self.sendAlerts(session, scom_ip, domain, username, password, idd.get("id"))
         return ids
 
-    def serviceTree(self, serviceidtree, session, scom_ip, domain, username, password, system_type):
+    def serviceTree(self, serviceidtree, session, scom_ip, domain, username, password, system_type,integration_mode):
```
lowdewijk, line 154: It is best to take all these variables `scom_ip, domain, username, password, system_type,integration_mode` and stick them in a single `ScomInstanceConfig` class. Then you can pass that around.

scom/stackstate_checks/scom/scom.py
```diff
@@ -157,6 +185,13 @@ def check(self, instance):
                 root_ids = []
                 root_tree = response.get("rows", [])
                 for component in root_tree:
+                    type_data = {
+                            "criteria": "(Id = '"+component.get("id")+"')"
+                            }
+                    type_response = (session.post('http://'+ scom_ip +'/OperationsManager/data/scomObjects',
+                                auth=HttpNtlmAuth(domain + '\\' + username, password),
+                                json=type_data["criteria"])).json()
+                    component_type = type_response.get("scopeDatas")[0].get("className").lower().replace(" ","-")
```
lowdewijk, line 194: what is this duplicate code?

scom/stackstate_checks/scom/scom.py
```diff
@@ -140,11 +166,13 @@ def check(self, instance):
         username = str(instance.get('username'))
         password = str(instance.get('password'))
         auth_method = str(instance.get('auth_mode'))
+        integration_mode = str(instance.get('integration_mode'))
```
lowdewijk, line 169: I would put the default to "api".

lowdewijk (commented):
In general:
 * Some tests would be nice.
 * Get rid of duplicate code
 * Please remain aware that in Python we use `snake_case` not `camelCase`.
 * Integration mode may not entirely be the right name, because it only works for events right now.

scom/stackstate_checks/scom/scom.py
melnemr: yaml.loads with json ? 
json.loads converts a string to json object. the string is for sure json because i convert it in powershell
powershell output is ASCII but can be set to utf-8 for example

scom/stackstate_checks/scom/scom.py
melnemr: indeed that the perfect scenario. However, there is a bug in stacksatate with stackpack url field contains "/" . if you enter "/" in the url field and the stackpack creates a view. You end up with totally empty view, even if you try to filter anything within this view, you do not get any components. That's why i am trying to avoid using "/" in the url

scom/stackstate_checks/scom/scom.py
melnemr: this was not complete, i have rewritten the whole integration and combined multiple operations together to save some requests.

scom/stackstate_checks/scom/scom.py
```diff
@@ -55,7 +56,24 @@ def get_instance_key(self, instance):
 
         instance_url = instance['hostip']
         return TopologyInstance(self.INSTANCE_TYPE, instance_url)
-
+    def sendAlertsPowershell(self,scom_server):
```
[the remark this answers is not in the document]
melnemr: fixed in the new code

melnemr:
New code added to new MR

PR #68, opened 2020-10-16
stac-10499: dynatrace topology integration, get config from core agent and duplic…

vishalxebia:
1. Dynatrace topology integration with tests.
2. Add method to get the config from agent.
3. Don't add duplicate components and relations in the snapshot.
4. Add method to get `conf.d` folder path.


dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+        self.log.info("Starting the collection of topology")
+        services = self.collect_services()
+        processes = self.collect_processes()
+        processes_groups = self.collect_process_groups()
+        hosts = self.collect_hosts()
+        applications = self.collect_applications()
+        topology = {"service": services, "process": processes, "process-group": processes_groups, "host": hosts,
+                    "application": applications}
+        # process component for each type
+        for comp_type, response in topology.items():
```
lowdewijk, line 77: Since you're only checking whether the response contains an error when processing, you might do a whole bunch of requests to DT before realizing that you've gotten an error. Perhaps you could check for the error in the `get_json_repsonse`? Perhaps call it `get_dynatrace_json_response` then.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+        Process each component type and map those with specific data
+        :param response: Response of each component type endpoint
+        :param component_type: Component type
+        :return: create the component on stackstate API
+        """
+        urn = ''
+        if "error" not in response:
+            for item in response:
+                # special case for host type as we get some float values
+                if component_type == "host":
```
lowdewijk, line 164: I'd make this a function calling `clean_item` and you should do this for all component types. Apparently any `float` OR `boolean` causes a problem for the agent. We should also drop this function asap, because it's just BS.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+                if component_type == "service":
+                    urn = "urn:service:/{}".format(externalId)
+                elif component_type == "process-group":
+                    urn = "urn:process-group:/{}".format(externalId)
+                elif component_type == "application":
+                    urn = "urn:application:/{}".format(externalId)
+                elif component_type == "process":
+                    urn = "urn:process:/{}".format(externalId)
+                elif component_type == "host":
+                    displayName = item.get("displayName")
```
lowdewijk, line 180: From the Dynatrace API I read that these are the fields that may contain hostnames:

```
esxiHostName
oneAgentCustomHostName
azureHostNames
publicHostName
localHostName
```

I would check them all and make all values you can find identifiers.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def collect_services(self):
+        """
+        Collects the response from the Dynatrace Service API endpoint
+        """
+        endpoint = self.url + "/api/v1/entity/services"
+        services = self.get_json_response(endpoint)
+        return services
+
+    def process_component(self, response, component_type):
```
lowdewijk, line 153: You are actually processing both components and relations, hence topology. `process_topology`?

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+                # special case for host type as we get some float values
+                if component_type == "host":
+                    for key in item.keys():
+                        if type(item[key]) is float:
+                            item[key] = str(item[key])
+                identifiers = []
+                data = dict()
+                externalId = item.get("entityId")
+                if component_type == "service":
+                    urn = "urn:service:/{}".format(externalId)
```
lowdewijk, line 172: I would not make this identifier not any of the other ones below except the host name one. I would simply add the dynatrace identifier generically like this:

```
external_id = item["entityId"] # let it fail if we can not find entityId
identifiers = list("urn:dynatrace:/{}".format(external_id))
if component_type = "host":
   ... # see comment below on where to get proper hostnames

File: dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py (hunk 3 above)
Comment (lowdewijk) [reply dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py:180]: Not sure whether `displayName` should be in that list.

File: dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py [hunk 6]
```diff
@@ -0,0 +1,276 @@
...
+                    urn = "urn:process:/{}".format(externalId)
+                elif component_type == "host":
+                    displayName = item.get("displayName")
+                    urn = "urn:host:/{}".format(displayName)
+                identifiers.append(urn)
+                # derive useful labels and get labels from dynatrace tags
+                labels = self.get_labels(item)
+                labels += self.get_labels_from_dynatrace_tags(item)
+                if "tags" in item:
+                    del item["tags"]
```
lowdewijk, line 187: Perhaps you should place this code in `filter_data`, then call it sooner and call it `filter_item_topology_data`.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+                # prefix the labels with `dynatrace-` for all labels
+                labels = ["dynatrace-{}".format(label) for label in labels]
```
lowdewijk, line 189: Do this at the end of `self.get_labels` for clarity.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+    def get_labels_from_dynatrace_tags(self, item):
+        """
+        Process each tag as a label in component
+        :param item: the component item to read from
+        :return: list of added tags as labels
+        """
+        tags = []
+        for tag in item.get("tags", []):
+            tag_label = ''
+            if bool(tag['context']) and tag.get('context') != 'CONTEXTLESS':
```
lowdewijk, line 214: `tag['context']` <- beware.. better use `tag.get('context')`

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+                elif component_type == "application":
+                    urn = "urn:application:/{}".format(externalId)
+                elif component_type == "process":
+                    urn = "urn:process:/{}".format(externalId)
+                elif component_type == "host":
+                    displayName = item.get("displayName")
+                    urn = "urn:host:/{}".format(displayName)
+                identifiers.append(urn)
+                # derive useful labels and get labels from dynatrace tags
+                labels = self.get_labels(item)
```
lowdewijk, line 184: About code: do what you say and say what you do. This doesn't get the labels, it gets _some_ of the labels...

You could simply call `get_labels_from_dynatrace_tags` from within `get_labels` to actually get all the labels.

dynatrace_topology/tests/samples/application_response_topology.json
```diff
@@ -0,0 +1,146 @@
...
+        ],
+        "instance": "https://ton48129.live.dynatrace.com",
+        "tags": [
+          "integration-type:dynatrace",
+          "integration-url:https://ton48129.live.dynatrace.com"
+        ],
+        "labels": [
+          "dynatrace-managementZones:allTypes",
+          "dynatrace-managementZones:app name exists",
+          "dynatrace-managementZones:Applications except easyTravel",
```
lowdewijk, line 40: Does stackstate support labels with spaces? Perhaps we need to sanitize label data for special characters?

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+            labels.append(item["entityId"])
+        if "monitoringState" in item:
+            actual_state = item["monitoringState"].get("actualMonitoringState")
+            expected_state = item["monitoringState"].get("expectedMonitoringState")
+            labels.append("actualMonitoringState:{}".format(actual_state))
+            labels.append("expectedMonitoringState:{}".format(expected_state))
+        if "softwareTechnologies" in item:
+            for technologies in item["softwareTechnologies"]:
+                tech_label = ''
+                if bool(technologies['type']):
```
lowdewijk, line 244: Why do you add this `bool` around it? That should do nothing right?  

Also, perhaps use `.get('type')` here and below instead? I don't think we should want to fail on this.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+        :return: None
+        """
+        outgoing_relations = component.get("fromRelationships", {})
+        incoming_relations = component.get("toRelationships", {})
+        for relation_type in outgoing_relations.keys():
+            # Ignore `isSiteOf` relation since location components are not processed right now
+            if relation_type != "isSiteOf":
+                for target_id in outgoing_relations[relation_type]:
+                    self.relation(externalId, target_id, relation_type, {})
+        for relation_type in incoming_relations.keys():
```
lowdewijk, line 97: If I believe Zandre, the agent is not going to help you dedup this. I also do not believe we can trust the batcher to not send a batch within the same check run, so the only thing we can do is make sure we have a unique `externalId` for the relation. This should ensure that StackState dedups the relations.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
vishalxebia: If we change to this then it will be very hard to merge the components from other data sources having the same name.  Because we had considered the same identifier for `host` type and in case of `service`, also followed the same way.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
zandrewitte: To add onto this one, if you are creating urn's import `Identifiers` from `stackstate_checks.base`. It has a bunch of functions to create the identifiers in the correct format. So if we ever need to change we just change it in one place. 

So it would be
```
Identifiers.create_host_identifier(hostname)
```

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+            # Ignore `isSiteOf` relation since location components are not processed right now
+            if relation_type != "isSiteOf":
+                for source_id in incoming_relations[relation_type]:
+                    self.relation(source_id, externalId, relation_type, {})
+
+    def filter_data(self, data):
```
zandrewitte, line 103: Can we make this a bit more generic and call it `unset`. Then we can just use it wherever we need with any key and it makes it really clear what we are removing. `del` mutates so it probably doesn't make sense to return data and assign it every time. So something like:

```
def unset(self, data, key):
  if key in data: del data[key]
```

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+        :param component_type: Component type
+        :return: create the component on stackstate API
+        """
+        urn = ''
+        if "error" not in response:
+            for item in response:
+                # special case for host type as we get some float values
+                if component_type == "host":
+                    for key in item.keys():
+                        if type(item[key]) is float:
```
zandrewitte, line 166: Are there more types we _might_ get here that will fail in the API? We can also add this to the base api and "sanitize" the data there?

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+                labels += self.get_labels_from_dynatrace_tags(item)
+                if "tags" in item:
+                    del item["tags"]
+                # prefix the labels with `dynatrace-` for all labels
+                labels = ["dynatrace-{}".format(label) for label in labels]
+                data.update(item)
+                data.update({
+                    "identifiers": identifiers,
+                    "tags": self.tags,
+                    "domain": self.domain,
```
zandrewitte, line 194: I think `self.domain` defaults to an empty string. The StackPack probably handles it fine, but let's not send data if it's not needed.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+            raise DynatraceError(response["error"].get("message"), response["error"].get("code"), component_type)
+
+    def get_labels_from_dynatrace_tags(self, item):
+        """
+        Process each tag as a label in component
+        :param item: the component item to read from
+        :return: list of added tags as labels
+        """
+        tags = []
+        for tag in item.get("tags", []):
```
zandrewitte, line 212: Is `tags` a dictionary? so we should default to `{}` then

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,276 @@
...
+        # append management zones in labels for each existing component
+        if "managementZones" in item:
+            for zone in item["managementZones"]:
+                labels.append("managementZones:{}".format(zone.get("name")))
+        if "entityId" in item:
+            labels.append(item["entityId"])
+        if "monitoringState" in item:
+            actual_state = item["monitoringState"].get("actualMonitoringState")
+            expected_state = item["monitoringState"].get("expectedMonitoringState")
+            labels.append("actualMonitoringState:{}".format(actual_state))
```
zandrewitte, line 239: `actual_state` and `expected_state` can be None right?

dynatrace_topology/tests/samples/application_response_topology.json
vishalxebia: it does as I saw this in AXA environment with spaces.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
vishalxebia: no it's a list.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
vishalxebia: we should probably either make some default value then.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
vishalxebia: good point I will add that and then can we make this in core agent as well?

stackstate_checks_base/stackstate_checks/base/stubs/topology.py
```diff
@@ -37,11 +37,20 @@ def _ensure_instance(self, check_id, instance_key):
             self._snapshots[check_id] = snapshot(instance_key)
         return self._snapshots[check_id]
 
+    def _is_duplicate_relation(self, check_id, instance_key, source_id, target_id, type, data):
+        return relation(source_id, target_id, type, data) in self._ensure_instance(check_id, instance_key)["relations"]
```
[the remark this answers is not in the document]
vishalxebia: removed.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
vishalxebia: i made it for now `filter_item_topology_data` to remove all three in the same method.

dynatrace_topology/README.md
```diff
@@ -0,0 +1,43 @@
...
+* _Applications_
+* _Processes_
+* _Process-Groups_
+* _Services_
+
+## Setup
+
+### Prerequisite
+
+* An API Token from Dynatrace must have access to read the smartscape topology.
```
lowdewijk, line 16: That token must have at least `DataExport`  token permission, I guess?
https://www.dynatrace.com/support/help/dynatrace-api/basics/dynatrace-api-authentication/

stackstate_checks_base/stackstate_checks/base/checks/base.py
```diff
@@ -212,7 +212,8 @@ def _raise_unexpected_type(self, argumentName, value, expected):
         raise ValueError("Got unexpected {} for argument {}, expected {}".format(type(value), argumentName, expected))
 
     def _check_struct_value(self, argumentName, value):
-        if value is None or isinstance(value, string_types) or isinstance(value, integer_types):
+        if value is None or isinstance(value, string_types) or isinstance(value, integer_types) or \
+                isinstance(value, float) or isinstance(value, bool):
```
lowdewijk, line 216: Great, you are fixing the problem where it sits! 

Do we still need `clean_unsupported_metadata` then?

lowdewijk (commented):
We're almost there! 👍

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,304 @@
...
+        if 'url' not in instance:
+            raise ConfigurationError('Missing URL in configuration.')
+        if 'token' not in instance:
+            raise ConfigurationError('Missing API Token in configuration.')
+
+        self.url = instance.get('url')
+        self.token = instance.get('token')
+        self.tags = instance.get('tags', [])
+        self.domain = instance.get('domain', self.url)
+        self.environment = instance.get('environment', 'production')
```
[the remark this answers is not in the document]
vishalxebia: @lowdewijk it's resolved properly in stackpack.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,304 @@
...
+        """
+        if 'url' not in instance:
+            raise ConfigurationError('Missing URL in configuration.')
+        if 'token' not in instance:
+            raise ConfigurationError('Missing API Token in configuration.')
+
+        self.url = instance.get('url')
+        self.token = instance.get('token')
+        self.tags = instance.get('tags', [])
+        self.domain = instance.get('domain', self.url)
```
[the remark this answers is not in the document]
lowdewijk: Does the StackPack generate views without knowing what the domain is?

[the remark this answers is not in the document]
vishalxebia: If a value is passed from here then it picks up otherwise if no value then just `Dynatrace`.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,304 @@
...
+        :param endpoint: Dynatrace API Endpoint to call
+        :param timeout: timeout for a response
+        :return: Response of each endpoint
+        """
+        headers = {"Authorization": "Api-Token {}".format(self.token)}
+        resp = None
+        msg = None
+        self.log.info("URL is {}".format(endpoint))
+        try:
+            session = Session()
```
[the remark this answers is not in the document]
vishalxebia: great I really missed this.

stackstate_checks_base/stackstate_checks/base/checks/base.py
vishalxebia: Yeah because this is integration only but in main core agent it still needs to be fixed.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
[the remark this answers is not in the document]
lowdewijk: Oh actually I see that the environment name is used for resolution with the common StackPack. This is all a bit tricky... Maybe we don't need to do anything about this, but at some point we may want to improve this in terms of clarity.

The relevant code:
```
"environments": [
--
  |   |   | \{{ getOrCreate (identifier "urn:stackpack:common" "Environment" element.data.environment) (identifier "urn:system:auto" "Environment" element.data.environment) (concat "Type=Environment;Name=" element.data.environment) \}}
  |   |   | ]
```

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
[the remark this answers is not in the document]
vishalxebia: If they want domain as `managementZones` then the data coming from dynatrace isn't actually of the same zone but others as well which they will confuse with.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,304 @@
...
+            # derive useful labels and get labels from dynatrace tags
+            labels = self.get_labels(item)
+            data.update(item)
+            self.filter_item_topology_data(data)
+            data.update({
+                "identifiers": identifiers,
+                "tags": self.tags,
+                "domain": self.domain,
+                "environment": self.environment,
+                "instance": self.url,
```
lowdewijk, line 222: A comment here that his url is used in StackState for component action may be helpful

dynatrace_topology/stackstate_checks/dynatrace_topology/data/conf.yaml.example
```diff
@@ -0,0 +1,33 @@
...
+    # verify: True  # By default it's True
+
+    # path to certificate file for https verification
+    # cert: /path/to/cert.pem
+
+    # path to public key of certificate for https verification
+    # keyfile: /path/to/key.pem
+
+    # domain to filter out different instances
+    # domain: "customer"
```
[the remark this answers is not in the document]
vishalxebia: agreed not to have mandatory but `dynatrace` as option default value.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,304 @@
...
+            session.verify = self.verify
+            if self.cert:
+                session.cert = (self.cert, self.keyfile)
+            resp = session.get(endpoint)
+            if resp.status_code != 200:
+                raise Exception("Got %s when hitting %s" % (resp.status_code, endpoint))
+        except requests.exceptions.Timeout:
+            msg = "{} seconds timeout when hitting {}".format(timeout, endpoint)
+            raise Exception("Exception occured for endpoint {0} with message: {1}".format(endpoint, msg))
+        return yaml.safe_load(resp.text)
```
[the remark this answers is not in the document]
vishalxebia: moved.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,304 @@
...
+        try:
+            with Session() as session:
+                session.headers.update(headers)
+                session.verify = self.verify
+                if self.cert:
+                    session.cert = (self.cert, self.keyfile)
+                resp = session.get(endpoint)
+            if resp.status_code != 200:
+                raise Exception("Got %s when hitting %s" % (resp.status_code, endpoint))
+            return yaml.safe_load(resp.text)
```
lowdewijk, line 301: Please move inside the `with`. This will ensure that if `resp.text` hits an exception that the session will be closed.

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -0,0 +1,304 @@
...
+        msg = None
+        self.log.debug("URL is {}".format(endpoint))
+        try:
+            with Session() as session:
+                session.headers.update(headers)
+                session.verify = self.verify
+                if self.cert:
+                    session.cert = (self.cert, self.keyfile)
+                resp = session.get(endpoint)
+            if resp.status_code != 200:
```
lowdewijk, line 299: indent once to the right

dynatrace_topology/stackstate_checks/dynatrace_topology/data/conf.yaml.example
```diff
@@ -0,0 +1,35 @@
...
+    # to verify the certificate for https
+    # verify: True  # By default it's True
+
+    # path to certificate file for https verification
+    # cert: /path/to/cert.pem
+
+    # path to public key of certificate for https verification
+    # keyfile: /path/to/key.pem
+
+    # domain to filter out different instances
```
lowdewijk, line 27: not to filter, but to use for the components

lowdewijk (approved):

PR #69, opened 2020-10-20
WIP: set batch_size default to 2500

hruhek:
https://stackstate.atlassian.net/browse/STAC-10855

More work on batching support

- [ ] Unit/Component test

### DIFF
--- a/servicenow/run_check.py
+++ b/servicenow/run_check.py
@@ -1,0 +1,1 @@
+/Users/hruhek/AG04/StackState/STAC-10855-servicenow-bug/run_check.py
--- a/servicenow/run_check.py
+++ b/servicenow/run_check.py
@@ -1,1 +1,0 @@
-/Users/hruhek/AG04/StackState/STAC-10855-servicenow-bug/run_check.py
--- a/servicenow/stackstate_checks/servicenow/servicenow.py
+++ b/servicenow/stackstate_checks/servicenow/servicenow.py
@@ -39,7 +39,10 @@ class ServicenowCheck(AgentCheck):
         auth = (user, password)
         batch_size = instance.get('batch_size', 2500)
         if batch_size > BATCH_MAX_SIZE:
-            raise ConfigurationError('Setting batch_size is {}. Max value is {}.'.format(batch_size, BATCH_MAX_SIZE))
+            raise ConfigurationError(
+                'Setting batch_size is: {} That is greater than max value of {}.'.format(batch_size, BATCH_MAX_SIZE)
+            )
+
         instance_tags = instance.get('tags', [])
         sys_class_filter = instance.get('include_resource_types', [])
 
--- a/servicenow/stackstate_checks/servicenow/servicenow.py
+++ b/servicenow/stackstate_checks/servicenow/servicenow.py
@@ -16,6 +16,9 @@ import requests
 from stackstate_checks.base import ConfigurationError, AgentCheck, TopologyInstance
 from stackstate_checks.base.errors import CheckException
 
+BATCH_DEFAULT_SIZE = 2500
+BATCH_MAX_SIZE = 10000
+
 
 class InstanceInfo:
     def __init__(self, instance_tags, base_url, auth, sys_class_filter, batch_size, timeout):
@@ -38,7 +41,9 @@ class ServicenowCheck(AgentCheck):
     def check(self, instance):
         base_url, password, user = self._get_mandatory_instance_values(instance)
         auth = (user, password)
-        batch_size = instance.get('batch_size', 1000)
+        batch_size = instance.get('batch_size', 2500)
+        if batch_size > BATCH_MAX_SIZE:
+            raise ConfigurationError('Setting batch_size is {}. Max value is {}.'.format(batch_size, BATCH_MAX_SIZE))
         instance_tags = instance.get('tags', [])
         sys_class_filter = instance.get('include_resource_types', [])
 
--- a/servicenow/stackstate_checks/servicenow/servicenow.py
+++ b/servicenow/stackstate_checks/servicenow/servicenow.py
@@ -2,12 +2,8 @@
 # All rights reserved
 # Licensed under a 3-clause BSD style license (see LICENSE)
 
-import json
-
-try:
-    json_parse_exception = json.decoder.JSONDecodeError
-except AttributeError:  # Python 2
-    json_parse_exception = ValueError
+import yaml
+from yaml.parser import ParserError
 
 # 3rd party
 import requests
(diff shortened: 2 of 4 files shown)


servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -221,6 +212,9 @@ def _collect_relations(self, instance_config, offset):
         return self._get_json_batch(url, offset, instance_config.batch_size, instance_config.timeout, auth)
 
     def _process_relations(self, relations, instance_config):
+        """
```
zandrewitte, line 215: If we're adding this comment, let's add some more detail about what it does / returns

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -244,37 +238,41 @@ def _get_json_batch(self, url, offset, batch_size, timeout, auth):
...
-                                   message=msg)
-                raise CheckException("Cannot connect to ServiceNow CMDB, please check your configuration.")
-
-        if resp.encoding is None:
-            resp.encoding = 'UTF8'
+    @staticmethod
+    def _get_json(url, timeout, auth=None, verify=True):
+        execution_time_exceeded_error_message = 'Transaction cancelled: maximum execution time exceeded'
+
+        response = requests.get(url, timeout=timeout, auth=auth, verify=verify)
```
zandrewitte, line 245: Why did this move out of a try, no longer catching the `requests.exceptions.Timeout`?

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -244,37 +238,41 @@ def _get_json_batch(self, url, offset, batch_size, timeout, auth):
...
-        if resp.get("error"):
-            raise CheckException("Problem in collecting CIs : {}".format(resp["error"].get("message")))
-        return resp
+        if response_json.get("error"):
+            raise CheckException(response_json["error"].get("message"))
+
+        return response_json
+
+    @staticmethod
+    def _get_mandatory_instance_values(instance):
```
zandrewitte, line 266: Instead of this why don't we use something like `schematics` to validate the models that we expect? Then we can validate the value and type for instance

hruhek: I was wondering if `schematics` project is still alive as last release on PyPI was in June 2018 and last commit to master December 2018. Then I found this https://github.com/schematics/schematics/issues/616
It seems that someone is still maintaining it.
OK, I'll introduce `schematics` to agent check.

servicenow/stackstate_checks/servicenow/servicenow.py
hruhek: We do in the main `check` method. All exception from here are handled there and they got critical service_check.
I added a test `test_get_json_timeout` that checks that.

servicenow/stackstate_checks/servicenow/servicenow.py
hruhek: @zandrewitte I used schematics to validate instance data. Please take a look if that is OK.
I would like to use it on a few more places like checking if result list exist in json response.

servicenow/stackstate_checks/servicenow/servicenow.py
hruhek: I'll fix all pydoc when I'm finished with other refactoring

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -174,62 +160,59 @@ def _process_components(self, components, instance_config):
                 identifiers.append("urn:host:/{}".format(comp_name))
             identifiers.append(external_id)
             data.update(component)
-            data.update({"identifiers": identifiers, "tags": instance_config.instance_tags})
+            data.update({"identifiers": identifiers, "tags": instance_info.instance_tags})
```
zandrewitte, line 163: I just noticed we still have this part here:
```
if 'fqdn' in component and component['fqdn']:
    identifiers.append("urn:host:/{}".format(component['fqdn']))	                
if 'host_name' in component and component['host_name']:
    identifiers.append("urn:host:/{}".format(component['host_name']))	               
else:
    identifiers.append("urn:host:/{}".format(comp_name))	             
```

Can you convert it to use the identifier functions like Dynatrace does: 
https://github.com/StackVista/stackstate-agent-integrations/blob/master/dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py#L5

https://github.com/StackVista/stackstate-agent-integrations/blob/master/dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py#L184

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -174,62 +160,59 @@ def _process_components(self, components, instance_config):
...
 
-        if "result" in state:
-            for relation in state.get('result', []):
-                id = relation['sys_id']
+        if "result" in types:
+            for relation in types.get('result', []):
+                sys_id = relation['sys_id']
                 parent_descriptor = relation['parent_descriptor']
-                relation_types[id] = parent_descriptor
+                relation_types[sys_id] = parent_descriptor
```
zandrewitte, line 187: I know it's nitpicking, but these can potentially fail if it's not there. We can introduce schematics here as well to avoid the `if x in relation` stuff. I would then propose to do it at a root level starting with `types`. One thing to note though, we want to stay resilient so we'll have to catch the validation exception, log it and continue with an empty relation_types dict.

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -174,62 +160,59 @@ def _process_components(self, components, instance_config):
...
-        return self._get_json_batch(url, offset, instance_config.batch_size, instance_config.timeout, auth)
+        return self._get_json_batch(url, offset, instance_info.batch_size, instance_info.timeout, auth)
 
-    def _process_relations(self, relations, instance_config):
-        relation_types = self._process_relation_types(instance_config)
+    def _process_relations(self, relations, instance_info):
+        """
+        process relations
+        """
+        relation_types = self._process_relation_types(instance_info)
```
zandrewitte, line 207: Same thing here. It can potentially fail if `parent` / `child` and the values aren't present. The lookup for `relation_types[type_sys_id]` can also fail.

servicenow/stackstate_checks/servicenow/servicenow.py
zandrewitte: You could argue that `_collect_relation_types` also just does the schematics validation and returns the list.

zandrewitte: Just remember to set `strict=False` when instantiating the schematics model like:
```
_event = Event(event, strict=False)  # strict=False ignores extra fields provided by nagios
_event.validate()
```
Then it will ignore extra fields and validation will be ok

zandrewitte (changes requested):

servicenow/run_check.py
```diff
@@ -0,0 +1 @@
+/Users/hruhek/AG04/StackState/STAC-10855-servicenow-bug/run_check.py
```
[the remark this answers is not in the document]
hruhek: Uf, that was a mistake. I'll deleted it from git.

servicenow/stackstate_checks/servicenow/servicenow.py
hruhek: Done.

PR #75, opened 2020-11-04
Stac-11097 reconnect on unauthenticated exception

vishalxebia:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/vsphere/stackstate_checks/vsphere/vsphere.py
+++ b/vsphere/stackstate_checks/vsphere/vsphere.py
@@ -551,7 +551,7 @@ class VSphereCheck(AgentCheck):
         try:
             self.server_instances[i_key].CurrentTime()
             self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK, tags=service_check_tags)
-        except Exception as e:
+        except Exception:
             # Try to reconnect, if the connection is definitely broken.
             self.server_instances[i_key] = self._smart_connect(instance, service_check_tags)
         return self.server_instances[i_key]
--- a/vsphere/stackstate_checks/vsphere/vsphere.py
+++ b/vsphere/stackstate_checks/vsphere/vsphere.py
@@ -378,6 +378,7 @@ class VSphereCheck(AgentCheck):
 
         # Connections open to vCenter instances
         self.server_instances = {}
+        self.session = None
 
         # Event configuration
         self.event_config = {}
@@ -488,14 +489,7 @@ class VSphereCheck(AgentCheck):
         now = time.time()
         return now - self.cache_times[i_key][entity][LAST] > self.cache_times[i_key][entity][INTERVAL]
 
-    def _get_server_instance(self, instance):
-        i_key = self._instance_key(instance)
-
-        service_check_tags = [
-            'vcenter_server:{0}'.format(instance.get('name')),
-            'vcenter_host:{0}'.format(instance.get('host')),
-        ]
-
+    def _smart_connect(self, instance, service_check_tags):
         # Check for ssl configs and generate an appropriate ssl context object
         ssl_verify = instance.get('ssl_verify', True)
         ssl_capath = instance.get('ssl_capath', None)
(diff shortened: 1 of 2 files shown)


vsphere/stackstate_checks/vsphere/vsphere.py
```diff
@@ -509,38 +503,57 @@ def _get_server_instance(self, instance):
...
+            )
+
+        try:
+            # Object returned by SmartConnect is a ServerInstance
+            # https://www.vmware.com/support/developer/vc-sdk/visdk2xpubs/ReferenceGuide/vim.ServiceInstance.html
+            server_instance = connect.SmartConnect(
+                host=instance.get('host'),
+                user=instance.get('username'),
+                pwd=instance.get('password'),
+                sslContext=context if not ssl_verify or ssl_capath else None,
```
zandrewitte, line 520: You dropped `connectionPoolTimeout=-1`, will we not run into the same issue if we make a long running requests for which this was the fix?

vsphere/stackstate_checks/vsphere/vsphere.py
```diff
@@ -509,38 +503,57 @@ def _get_server_instance(self, instance):
...
-                self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL,
-                                   tags=service_check_tags, message=err_msg)
-                raise Exception(err_msg)
-            self.server_instances[i_key] = server_instance
+            self.server_instances[i_key] = self._smart_connect(instance, service_check_tags)
 
         # Test if the connection is working
         try:
-            self.server_instances[i_key].RetrieveContent()
+            self.server_instances[i_key].CurrentTime()
```
zandrewitte, line 552: `server_instance.CurrentTime()` is already done in the `_smart_connect` function, so we don't have to do it again here

vsphere/stackstate_checks/vsphere/vsphere.py
```diff
@@ -509,38 +503,57 @@ def _get_server_instance(self, instance):
...
             self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.OK, tags=service_check_tags)
-        except Exception as e:
-            err_msg = "Connection to %s died unexpectedly: %s" % (instance.get('host'), e)
-            self.service_check(self.SERVICE_CHECK_NAME, AgentCheck.CRITICAL,
-                               tags=service_check_tags, message=err_msg)
-            raise Exception(err_msg)
+        except Exception:
+            # Try to reconnect, if the connection is definitely broken.
+            self.server_instances[i_key] = self._smart_connect(instance, service_check_tags)
```
zandrewitte, line 556: If this reconnect also fails we should drop `i_key` from the `self.server_instances` map, otherwise we keep the "broken" connection in the state

zandrewitte (changes requested):

vsphere/stackstate_checks/vsphere/vsphere.py
vishalxebia: @zandrewitte we are overriding it anyway, so it won't matter I think.

vsphere/stackstate_checks/vsphere/vsphere.py
vishalxebia: @zandrewitte Yeah, I also thought the same but the scenario could be like if we already have `i_key` in the dictionary then we are not going into that loop. And when we do connect the first time then we also test if we have enough permission to make calls.

vsphere/stackstate_checks/vsphere/vsphere.py
vishalxebia: looks like not, the main issue with this kind of problem was only `session` timing out from vCenter Esxi and only way to fix this is to re-login and this is what the fix made right now.

vsphere/stackstate_checks/vsphere/vsphere.py
zandrewitte: Ah right of course, ok resolving this one

vsphere/stackstate_checks/vsphere/vsphere.py
zandrewitte: If `self._smart_connect(instance, service_check_tags)` fails and throws an exception on either the `connect.SmartConnect` or `server_instance.CurrentTime()` we won't override it and that connection will remain in `self.server_instances`. So next time the check runs it will fail again on  `self.server_instances[i_key].CurrentTime()`.

Does that make sense? It won't blow up, but it will fail once unnecessarily

vsphere/stackstate_checks/vsphere/vsphere.py
zandrewitte: Ok just verify this. I remember you did some test scenario last time round to verify this fix worked for the previous customer. Try and run that again and see what happens

vsphere/stackstate_checks/vsphere/vsphere.py
vishalxebia: ok. done.

vsphere/stackstate_checks/vsphere/vsphere.py
vishalxebia: @zandrewitte yeah, I know but the other customer is also facing this now I think.

zandrewitte (approved):

PR #84, opened 2020-11-25
Adding removal of empty elements

hruhek:
https://stackstate.atlassian.net/browse/STAC-10665

_fix_encoding in base now removes empty elements

- [x] Unit/Component test

### DIFF
--- a/stackstate_checks_base/stackstate_checks/base/checks/base.py
+++ b/stackstate_checks_base/stackstate_checks/base/checks/base.py
@@ -393,9 +393,9 @@ class AgentCheckBase(object):
     def component(self, id, type, data, streams=None, checks=None):
         instance = self._get_instance_key_value()
         try:
-            fixed_data = self._fix_encoding(data)
-            fixed_streams = self._fix_encoding(streams)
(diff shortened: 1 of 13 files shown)


stackstate_checks_base/stackstate_checks/base/checks/base.py
```diff
@@ -667,28 +669,48 @@ def _get_requests_proxy(self):
...
+            encoding_list = [element for element in list(field) if self._is_field_empty(element)]
             for i, element in enumerate(encoding_list):
-                encoding_list[i] = self._fix_encoding(element, "element of set")
-            value = set(encoding_list)
-        return value
+                encoding_list[i] = self._sanitize(element, "element of set")
+            field = set(encoding_list)
+        return field
+
+    def _is_field_empty(self, field):
```
TizianoPerrucci, line 703: maybe I'm too tired but shouldn't this function be called the **opposite**, thus `_is_not_empty` ?

because only if not empty we will retain it during sanitization

TizianoPerrucci (changes requested):

PR #86, opened 2020-11-26
Stac 11444 no yaml dynatrace

TizianoPerrucci:
https://stackstate.atlassian.net/browse/STAC-11444

It replaces `yaml.safe_load()` with `json.load()` which should prevent memory issues in the future.
It add unicode characters in all the canned responses to ensure unit tests can deal with it properly everywhere.
It replaces also `.format()` in favor of `%` for concatenating strings because the `.format()` returns a string and when dealing with unicode will it fails, instead the `%` operator returns always a unicode.

- [x] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/dynatrace_topology/requirements-dev.txt
+++ b/dynatrace_topology/requirements-dev.txt
@@ -1,1 +1,2 @@
 -e ../stackstate_checks_dev
+requests_mock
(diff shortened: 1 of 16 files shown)


hruhek (approved):
I have one minor concern. GitHub UI doesn't shows the Unicode character we used. It's invisible, and that confused that me for a bit unit I did checkout of the branch and inspected the files on my machine. Maybe we could use some visible Unicode character.

I'll approve it, your change works and this is really a minor issue.

TizianoPerrucci:
@hruhek sure, the reason why I picked `\u0081` is because we have it mentioned in this bug https://stackstate.atlassian.net/browse/STAC-11294

PR #87, opened 2020-11-27
Stac 11357 snow remove str.format it causes py2 utf-8 encoding problems

hruhek:
https://stackstate.atlassian.net/browse/STAC-11357

`str.format` causes unicode problems in py2 so we are switching to `%` operator for string concatenation

### DIFF
--- a/servicenow/stackstate_checks/servicenow/servicenow.py
+++ b/servicenow/stackstate_checks/servicenow/servicenow.py
@@ -314,6 +314,7 @@ class ServicenowCheck(AgentCheck):
 
     def _process_change_requests(self, instance_info):
         response = self._collect_change_requests(instance_info)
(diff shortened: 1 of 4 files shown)


TizianoPerrucci (changes requested):
Most of the string concatenation are in creating nice error msg when throwing exceptions, or when creating the servicenow query url; although is good to cover them I don't think we should be really worried about them.

The one that really is concerning are the concatenation happening when creating the topology events. Can we introduce few unicode characters in the test canned response `CHG0000001.json` so we avoid introducing regression in the future ?

hruhek (commented):
@vishalxebia I cannot approve this as I'm the author of this PR.

vishalxebia (approved):

PR #98, opened 2020-12-22
STAC-11469 remove yaml safe load from vsphere

hruhek:
https://stackstate.atlassian.net/browse/STAC-11469

Removed `yaml.safe_load(json.dumps(data_from_vsphere))` as we do cleanup in `base.py` with `_fix_encoding`.

- [x] Unit/Component test


hruhek:
We integrated vSphere upstream changes from DataDog, so this PR is not needed any more.

PR #108, opened 2021-03-02
stac-10915 servicenow tags

vishalxebia:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/servicenow/stackstate_checks/servicenow/servicenow.py
+++ b/servicenow/stackstate_checks/servicenow/servicenow.py
@@ -173,7 +173,7 @@ class ServicenowCheck(AgentCheck):
         return sysparm_query
 
     @staticmethod
-    def _apply_field_metadata(data, display_value_list):
+    def select_metadata_field(data, display_value_list):
         """
         Retrieve the proper attribute either `display_value` or `value` from data
         :param data: metadata from servicenow
@@ -248,10 +248,11 @@ class ServicenowCheck(AgentCheck):
                 config_item = ConfigurationItem(component, strict=False)
                 config_item.validate()
             except DataError as e:
-                self.log.warning("Error while processing properties - {}".format(e))
+                self.log.warning("Error while processing properties of CI {} having sys_id {} - {}"
+                                 .format(config_item.sys_id.value, config_item.name.value, e))
                 continue
             data = {}
-            component = self._apply_field_metadata(component, instance_info.component_display_value_list)
+            component = self.select_metadata_field(component, instance_info.component_display_value_list)
             identifiers = []
             comp_name = config_item.name.value
             comp_type = config_item.sys_class_name.value
@@ -298,10 +299,10 @@ class ServicenowCheck(AgentCheck):
                 ci_relation = CIRelation(relation, strict=False)
                 ci_relation.validate()
             except DataError as e:
-                self.log.warning("Error while processing properties - {}".format(e))
+                self.log.warning("Error while processing properties of relation having Sys ID {} - {}".format(ci_relation.sys_id.value, e))
                 continue
             data = {}
-            relation = self._apply_field_metadata(relation, instance_info.relation_display_value_list)
+            relation = self.select_metadata_field(relation, instance_info.relation_display_value_list)
             parent_sys_id = ci_relation.parent.value
             child_sys_id = ci_relation.child.value
             # first part after splitting with :: contains actual relation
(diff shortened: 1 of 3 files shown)


servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -57,6 +62,27 @@ class ChangeRequest(Model):
...
+    name = ModelType(WrapperStringType, required=True)
+    sys_class_name = ModelType(WrapperStringType, required=True)
+    sys_id = ModelType(WrapperStringType, required=True)
+    sys_tags = ModelType(WrapperStringType, default=WrapperStringType())
+    fqdn = ModelType(WrapperStringType, default=WrapperStringType())
+    host_name = ModelType(WrapperStringType, default=WrapperStringType())
+
+
+class CIRelation(Model):
+    sys_id = ModelType(WrapperStringType)
```
zandrewitte, line 75: Is `sys_id` required?

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -209,52 +244,37 @@ def _process_components(self, instance_info):
         collected_components = self._batch_collect(self._batch_collect_components, instance_info)
 
         for component in collected_components:
+            try:
+                config_item = ConfigurationItem(component, strict=False)
+                config_item.validate()
+            except DataError as e:
+                self.log.warning("Error while processing properties - {}".format(e))
```
zandrewitte, line 251: Add the component `sys_id` and `name` to the log

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -145,17 +173,24 @@ def _get_sys_class_relation_filter_query(self, sys_class_filter):
         return sysparm_query
 
     @staticmethod
-    def _filter_empty_metadata(data):
+    def _apply_field_metadata(data, display_value_list):
```
zandrewitte, line 176: This actually only selects `display_value` or `value` depending on whether it's in `display_value_list`. Can we maybe rename the function to `select_metadata_field`... or something better :smile:

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -272,16 +292,28 @@ def _process_relations(self, instance_info):
...
-            type_sys_id = relation['type']['value']
-
-            relation_type = relation_types[type_sys_id]
-            data = self._filter_empty_metadata(relation)
-            data.update({"tags": instance_info.instance_tags})
+            try:
+                ci_relation = CIRelation(relation, strict=False)
+                ci_relation.validate()
+            except DataError as e:
+                self.log.warning("Error while processing properties - {}".format(e))
```
zandrewitte, line 301: Add the relation `sys_id` ro the log

servicenow/stackstate_checks/servicenow/servicenow.py
```diff
@@ -272,16 +292,28 @@ def _process_relations(self, instance_info):
...
+                continue
+            data = {}
+            relation = self._apply_field_metadata(relation, instance_info.relation_display_value_list)
+            parent_sys_id = ci_relation.parent.value
+            child_sys_id = ci_relation.child.value
+            # first part after splitting with :: contains actual relation
+            relation_type = ci_relation.type.display_value.split("::")[0]
```
zandrewitte, line 308: Let's maybe add a check here to make sure the display value is not empty and that it contains a `::`, so something like:
```
if ci_relation.type.display_value and "::" in ci_relation.type.display_value
```

zandrewitte (changes requested):

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: nope it's optional. we just need `parent id` and `child id`. We don't have any concept for `external_id` right now for relation. so not required.

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: yeah, I couldn't think better.

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: it's actually a required field and we do that using schematics already. should we validate again?

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: added.

servicenow/stackstate_checks/servicenow/servicenow.py
vishalxebia: added.

zandrewitte (approved):

PR #110, opened 2021-03-09
Stac 12061 http handler

MelcomVE:
https://stackstate.atlassian.net/secure/RapidBoard.jspa?rapidView=98&projectKey=STAC&modal=detail&selectedIssue=STAC-12061

- Create a common base class that assists in doing HTTP requests

- [x] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,653 @@
...
+
+    """
+        This function is used to send out the request that has been build up
+        It has one of two return types
+            - The response from the request
+            - Or None if the response fails any validations unless no validation was defined and the response is JSON
+    """
+    def send(self):
+        # Send the request out
+        response = self.session.send(self.request_object.prepare(),
```
lowdewijk, line 552: It would be better separation of concerns to let the caller create the request object before and then hand that to this method as a parameter. Otherwise you get a method on a class that you can only call under certain circumstances and you now first have to figure out a bunch of prerequisites before you can use it. In general we should strive for as much referential transparency as possible.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,653 @@
...
+        else:
+            return self.request_object.auth
+
+    """
+        Set a proxy that the request will use
+
+        @proxies
+            The proxy object as defined in the requests library
+    """
+    def set_proxy(self, proxies=None):
```
lowdewijk, line 430: I would separate out the connection configuration into a separate class and then pass that to the HttpHelper. This should also help us generify the actual yaml configuration and is better separation of concerns.

lowdewijk (commented):
In general it looks really clean and has all the right characteristics, excepting the unicode handling. 

However I would advise splitting `HttpHelper` up. It does not really adhere to the single responsibility principle. Right now `HttpHelper` is doing at least three things:

1. Connection configuration 
2. Request building 
3. Session handling
4. Executing requests against the connection

I can imagine that you'd want to bunch 3 and 4 together, as the session can be transparent to the user, but I'd definitely split out 1 and 2.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,653 @@
...
+        'body': False
+    })
+
+
+"""
+    Authentication Models
+"""
+
+
+class _HTTPBasicAuth(Model):
```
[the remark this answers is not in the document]
MelcomVE: Added this to show that this class is for internal use, HTTPAuthenticationType references that class as a Basic Auth model

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,1375 @@
...
+    hello = StringType(required=True)
+    pong = BooleanType(required=True)
+
+
+"""
+ HTTP Helper Common Class
+"""
+
+
+class TestHTTPHelperCommon(unittest.TestCase):
```
lowdewijk, line 39: please also write a test with some weird unicode characters in both the request and the response.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+
+
+class HTTPBasicAuthentication(Model):
+    username = StringType(required=True)
+    password = StringType(required=True)
+
+
+"""
+    Authentication ENUM
+"""
```
lowdewijk, line 30: I am a bit puzzled by your docstrings. AFAIK you have to put the docstring underneath the class/function declaration.  E.g.

```
class HTTPAuthenticationType(Enum):
    """
    Authentication ENUM
    """
```

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+
+try:
+    import urlparse
+except ImportError:
+    import urllib.parse as urlparse
+
+
+"""
+    Authentication Models
+"""
```
lowdewijk, line 20: Please remove useless comments. No need to repeat the obvious.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+    The HTTP Helper Common class is used for common functionality split between all the other
+    HTTP Helper classes.
+    Functionality:
+        - Mapped enums to lists
+        - Log messages
+        - Create a dictionary from a string
+"""
+
+
+class HTTPHelperCommon:
```
lowdewijk, line 69: This class is a weird mix of stuff.

Methods that are static can only just be functions.
Parameters that are constant can also just be module scoped variables.

I would rename this thing to printer, because that's the only real purpose it serves.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+    _verbose = False
+
+    # Mapped enum values
+    http_method_enums = [item.value for item in HTTPMethod]
```
lowdewijk, line 73: why not place this on HTTP method itself? You can use class variables.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+    @staticmethod
+    def print_error(error):
+        raise Exception(error)
+
+    @staticmethod
+    def print_not_implemented_error(error):
+        raise NotImplementedError(error)
+
+    @staticmethod
+    def print_type_error(error):
```
lowdewijk, line 96: if you say `print` I will expect `print`, not raise. This breaks the POLA (https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+      Thus if the developer use the setters and getter those internal state will be manipulated
+"""
+
+
+class HTTPHelper:
+    _common = HTTPHelperCommon()
+    _connection = HTTPHelperConnectionHandler()
+    _request = HTTPHelperRequestHandler()
+    _session = HTTPHelperSessionHandler()
+    _response = HTTPHelperResponseHandler()
```
lowdewijk, line 1298: Why do you create these class variables?

lowdewijk (commented):
In general it looks much better. The decoupling is much better and the error messages are great!

But now it has exploded in lines of code :)

If I look at this code I see a lot of lines of code that deal with getting, setting, clearing, but I wonder why.

Why do we need the option to change the proxy of a connection? Why do we need the option to clear the request headers? Can't we just construct a new request on each request? Why do we need all that state to stick around?

If the pattern is to use the helper like this:

```
        request = HTTPHelper(self.verbose)
        response = request.get(
            request_schematic_validation=BodyRequestSchematicTest,
            request_type_validation=HTTPRequestType.JSON,
            mock=True,
            mock_status=200,
            mock_response={'hello': 'world'},
            url="mock://www.google.com",
            body={'hello': 'world', 'pong': True},
            headers={'header': 'a'},
            query={'query': 'b'},
            timeout=30,
            retry_policy=dict(
                total=3
            ),
        )
```

Then why do we need all these mutable and stateful objects? Why do we need the ability to mutate all variables when you can just recreate them? Why do these objects need to keep a requests-lib object under the hood (i.e. keep state) always synced up and ready to go? Can't we just create the requests-lib's object just on the fly when doing the send?

When you think of it: the request handler class could also be a single simple pure (referentially transparent) function: you give it a dict with all information that is needed for a request and it gives you a requests-lib object back. Plain and simple. Is there a reason to make it more complicated than that? 

A lot of code is also spent on type checking. I would love for us to use a type system, but that is beyond the scope of this PR.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+        - SET && GET HTTP Timeout
+        - SET && GET HTTP Retry Policy
+        - SET && GET HTTP SSL Verification
+        - SET && GET HTTP Proxy
+        - Sending the HTTP Request and Session to the Endpoint
+"""
+
+
+class HTTPHelperConnectionHandler:
+    # Private objects to keep internal state
```
lowdewijk, line 832: You are defining class variables here, not creating private properties.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+
+            elif isinstance(self._validate_type.value, type(response.content)) is False:
+                errors.append("""The response content type does not conform to the validation type. The response
+                                 type is {0}.
+                                 The expected type is {1}
+                                 To fix this you can either modify the type validation or remove it entirely
+                                 """.format(str(type(response.content)), str(self._validate_type.value)))
+
+        # Return all the errors that occurred
+        return errors
```
lowdewijk, line 1264: In all other places you raise exceptions, but all of a sudden here you return errors? I would say that by this point you might as well just raise an exception here with an exception with a message that contains all errors.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+      These can then manually be passed down as kwargs to the send function
+    - The developer can create the HTTPHelperRequestHandler, HTTPHelperSessionHandler, HTTPHelperResponseHandler and
+      HTTPHelperConnectionHandler.
+      These can then manually be set within the HTTPHelper class with the overwrite functions
+    - Or it can be left to the internal state where the Helper control and build the state.
+      Thus if the developer use the setters and getter those internal state will be manipulated
+"""
+
+
+class HTTPHelper:
```
lowdewijk, line 1293: I would place this at the top of the file. This is what you want people to use and read about.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+    authentication_types = [item.value for item in HTTPAuthenticationType]
+    response_types = [item.value for item in HTTPResponseType]
+
+    def __init__(self, verbose=False):
+        self._verbose = verbose
+
+    # Print verbose message to track events
+    def print_verbose(self, message):
+        if self._verbose is True:
+            print(message)
```
lowdewijk, line 84: we must have a proper logger available right?

lowdewijk: we should use a logger and use that directly, not a helper common class that prints to stdout.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
lowdewijk: Scrap that. We should use a real logger.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+        passed to the function 'http://www.url.com?test=123' the query parameters on this URL will be split out and
+        applied into the correct object within the Request() object
+
+    Input:
+        @url
+            A string object containing the endpoint that should be queried
+    """
+
+    def set_url(self, url):
+        self._common.print_verbose("Attempting to set the active current url to {0}".format(str(url)))
```
lowdewijk, line 275: You know. All these log statements, could be drastically simplified if you'd just get a dict with all the information in it that you need to transform to a requests-lib function. You'd then simply log the dict and be finished.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+    """
+    Functionality:
+        Lorem Ipsum
+
+    Input:
+        @parameters
+            Lorem Ipsum
+    """
+
+    def set_query_param(self, parameters):
```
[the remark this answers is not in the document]
MelcomVE: Zandre requested that the library should support request and session objects, this is to persist state between multiple requests

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+
+    def validate(self, response):
+        self._common.print_verbose("Attempting to validate the response.")
+
+        errors = list([])
+
+        # If the HTTP Response Validation Schematic has been set
+        if self._validate_schematic is not None:
+            try:
+                print("awe")
```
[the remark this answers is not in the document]
MelcomVE: Oops :D 🗡️

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Sorry I'm use to defining information above the starting line for the defined information

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: With the previous version you mentioned to not throw on the response and rather allow the developer to handle it. With the above I am compile a list of things that went wrong instead of throwing a error

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Sure will do

stackstate_checks_base/tests/test_http_helper.py
MelcomVE: Will do

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Will do

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: I'll do some investigation to see if there is a existing logger being used within the other services

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Sure, I'll move those VARIABLES

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Might replace this if I find a proper logger being used within the other services

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: This is to allow something like a simple GET or POST function. Those objects needs to be created so that the send function can use them. This makes it easier for the developer to not be forced to manually create them and pass them into the GET function but rather just call the GET function.
There is also functionality to allow the developer to manually overwrite anyone of those if need be.

If these are removed then the developer will always be expected to pass these as parameters, If this is a beter approach then I will gladly remove those constructors.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: I'm referring to the _ variables inside the class.
Redundant comment will remove them

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Might replace this if I find a proper logger being used within the other services

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: The intent for those verbose prints was to indicate a single object before it was applied. This will allow to see what was the action before a possible error.

The dict object might give a bigger picture but not a precise action before a error occured.

When I complete the following comment "Instead of always keeping the request object in sync at all places, you can also just construct the request object on demand through a method. This is safer and easier to follow." I will include the dict inside the logs to indicate the current state and the attempted action

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1447 @@
...
+            self._common.print_verbose("Endpoint parsed successfully, Result is as follow {0}".format(str(parsed_url)))
+
+            # The URL is recreated here by adding the schema, net location and path together
+            url = (parsed_url.scheme + "://" if len(parsed_url.scheme) > 0 else "") \
+                + parsed_url.netloc \
+                + parsed_url.path
+            self._common.print_verbose("Reconstructed URL {0}".format(str(url)))
+
+            # Apply the reconstructed Endpoint to the Request() object
+            self._request.url = url
```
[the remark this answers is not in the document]
MelcomVE: Will do, I'll keep dict state inside the request and session objecsts and only apply the values on demand

MelcomVE:
Why do we need the option to change the proxy of a connection? 
	- I added this to add to the HTTP helper class functionality. If it seems that this is a bit unnecessary then we can remove it.

Why do we need the option to clear the request headers? 
	- If a request object is reused this assists in removing data. I can remove the clear functions forcing the correct data to be set once and if something needs to be removed then the request object must be recreated.
	- A session object will persist tho, If a session is reused without a clear function that will force you to use the prev state as you can not overwrite them only join to it.

Can't we just construct a new request on each request? 
	- I'll need to chat to Zandre to find out if this is fine, We can then force a new request to be constructed after the data has been send

Why do we need all that state to stick around?
Then why do we need all these mutable and stateful objects? 
	- This allows for the reusability of requests. We are applying the values received to an object allowing the developer to change it at a later stage if required without needing to recreate everything.

Why do we need the ability to mutate all variables when you can just recreate them? 
	- Not sure I fully understand this one. If we apply a variable we need to store it somewhere. Need it be in the Request() object or a dict thus updating one will overwrite that value again

Why do these objects need to keep a requests-lib object under the hood (i.e. keep state) always synced up and ready to go?
	- I can convert the Request() and Session() object to an internal dict but will serve almost the same purpose as those 2 objects are also just storing data until executed.
	- Altho as mentioned a dict will keep it clean as the Request() and Session() construction is happening in one place within the send functionality
 
Can't we just create the requests-lib's object just on the fly when doing the send?
	- I will convert the request and session helpers to use a dict and only create the request and session object on the send function

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Baked the creation of these class variables into the builder function. The builder function will now create these, apply the correct data and trigger the send. This also reduced mutability

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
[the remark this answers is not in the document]
MelcomVE: Removed query parameters from the Session as it makes sense to keep that type of object on the Request object. If there is a special case where it is more convenient to use a session then we can bring it back

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Got removed with the updated python logging lib implementation

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Move to within the classes

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Got removed with the updated python logging lib implementation

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Got removed with the updated python logging lib implementation

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+        if self._connection_model.proxy is not None:
+            session.proxies = self._connection_model.proxy
+
+        # Apply the ssl verification
+        if self._connection_model.ssl_verify is not None:
+            session.verify = self._connection_model.ssl_verify
+
+        # Auth will already have gone through validation thus we just test types and apply values
+        # Request - HTTPBasicAuth
+        if request_handler.get_request_model().auth_type is HTTPAuthenticationType.BasicAuth:
```
lowdewijk, line 1004: why doesn't `create_request_object` do this?

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+            request.auth = HTTPBasicAuth(request_handler.auth_data.get('username'),
+                                         request_handler.auth_data.get('password'))
+
+        # Session - HTTPBasicAuth
+        if session_handler.get_session_model().auth_type is HTTPAuthenticationType.BasicAuth:
```
lowdewijk, line 1009: why doesn't `create_session_object` do this?

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+class HTTPRequestType(Enum):
+    JSON = dict
+
+
+class HTTPResponseType(Enum):
+    PLAIN = str
+    JSON = dict
+
+
+class HTTPHelperModel(Model):
```
lowdewijk, line 45: It's good that you didn't get tempted to pass this model down to all classes. That would've saved you some code, but would have made entangled all your code and would've made it hard to disentangle at some later stage when you want to use/test things in isolation 👍

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+        request
+    """
+    mock_enable = BooleanType(required=False, default=False)
+    mock_status = IntType(required=False, default=None)
+    mock_response = BaseType(required=False, default=None)
+    endpoint = StringType(required=True, default=None)
+    auth_data = DictType(BaseType, required=False, default=None)
+    auth_type = BaseType(required=False, default=None)
+    session_auth_data = BaseType(required=False, default=None)
+    session_auth_type = BaseType(required=False, default=None)
```
lowdewijk, line 58: I wonder what's the point of having session be part of this model. If you use this model you only get the response back, but you won't actually get the session back, therefore you can't actually use session this way. 

You should consider how a user of this lib is going to use sessions.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+            - PUT Request
+
+        All the HTTPHelper classes and Sub classes can accept a schematic dictionary that allows certain values to
+        manipulate state with
+    """
+
+    def __init__(self):
+        self.log = logging.getLogger('%s.base' % __name__)
+
+    def get(self, http_model):
```
lowdewijk, line 92: Perhaps it would be good to add another parameter here to pass a session handler for reuse purposes. Same for the other methods in this class.

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,458 @@
...
+
+        assert response.request.method == "GET"
+        assert response.status_code == 201
+        assert json.loads(response.content.decode(response.encoding)).get("hello") == "world"
+        assert response.request.url == "mock://test.com"
+
+    def test_full_unicode_request_and_response(self):
+        unicode = ""
+        for i in range(100):
+            unicode = unicode + chr(i)
```
lowdewijk, line 409: How is this testing unicode? You are just setting `unicode` to a string containing the first 100 ASCII characters. You are not even getting outside the range of 7 bit ASCII this way.

Pleas read this: https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/

The bare minimum you should test for is receiving responses that contain some weird Chinese characters in a non UTF-8 encoding, which you should be able to decode properly by reading the `Content-Type` HTTP header. And by you I actually mean requests; that should all be taken care of by requests if all goes well.

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,458 @@
...
+            "request_schematic_validation": BodyRequestSchematicTest,
+            "request_type_validation": HTTPRequestType.JSON,
+            "response_status_code_validation": 200,
+            "response_type_validation": HTTPResponseType.JSON,
+            "response_schematic_validation": BodyResponseSchematicTest,
+        })
+
+        assert response.request.method == "GET"
+        assert response.status_code == 200
+        assert json.loads(response.content.decode(response.encoding)).get("hello") == "world"
```
lowdewijk, line 383: I would expect `json.loads(response.content.decode(response.encoding)` this code to be taken care of by the response handler. You are now leaking the `requests` library to the user, which you actually meant to wrap. This now forms a leaky abstraction. 

My advice: do not provide the actual response back to the user, but a wrapper around it that you control. You can simply have two methods on that wrapper like `get_json` and `get_text`. 

You can even call the validate on the response only when the user calls `response.get_json()` and not trigger it on `response.get_text()`.

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,458 @@
...
+    def test_get_request_model(self):
+        request_handler = HTTPHelperRequestHandler()
+        model = request_handler.get_request_model()
+        assert isinstance(model, HTTPHelperRequestModel)
+
+        # Test the base model. This should fail as a few things are required
+        try:
+            request_handler.validate_request_model()
+        except DataError as e:
+            assert len(e.errors) == 2
```
lowdewijk, line 44: You are actually meant to use these assert methods. https://docs.python.org/3/library/unittest.html#assert-methods 

https://stackoverflow.com/a/45947566/1860591

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,458 @@
...
+
+        try:
+            HTTPHelperRequestHandler().validate_endpoint()
+            assert False
+        except ValueError:
+            assert True
+
+        try:
+            HTTPHelperRequestHandler(-1).validate_endpoint()
+            assert False
```
lowdewijk, line 83: https://docs.pytest.org/en/stable/reference.html#pytest-fail is how you fail a test. `assert False` will just produce a meaningless result. Writing a failure message gives you a chance to explain to the reader of both your code and the failed test what you had expected.

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,458 @@
...
+        assert HTTPHelperRequestHandler({"request_schematic_validation": -1}) \
+               .get_request_model().request_schematic_validation == -1
+        assert HTTPHelperRequestHandler({
+            "body": {
+                "hello": "world",
+                "pong": True
+            },
+            "request_schematic_validation": BodyRequestSchematicTest
+        }).validate_body_schematic()
+        try:
```
lowdewijk, line 160: Typically unit test frameworks allow you to test for the presence of exceptions. 

pytest is no exception: https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions 

pun intended :)

lowdewijk (commented):
This is again a major leap forward. You're almost there :)

A lot of the comments I've made apply to a lot of code. I didn't repeat these comments each time I saw the same issue (e.g. misappropriation of class variables, useless setters, etc). 

Think about the previous two iteration: 
 * The previous iteration did the same as this, but all the type checks were manual and there was a lot of getting, setting, clearing, logging that took up a whole lot of space and provided ample room for error. Since the previous iteration you've reduced over 200 lines of code on the helper, but once you get rid of all those setters and follow some of my comments you will have reduce it by another 200. 
 *  The iteration before this everything happened in a single class and there were a _lot_ of if statements. I am sure it would have been a nightmare to deal with the bugs that you would have found in this.

In general I hope this is a good learning experience for you and you can keep up this type of quality in the future at StackState; our customers need it for monitoring their IT environment.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+        self.validate_body()
+        request_object.data = self._request_model.body
+
+        # Authentication Apply and validation
+        request_object.auth_data = self._request_model.auth_data
+        request_object.auth_type = self._request_model.auth_type
+
+        return request_object
+
+    def _set_method(self, method):
```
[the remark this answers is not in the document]
MelcomVE: Forgot about them they basically do nothing as it private and not used removed it

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+
+        else:
+            message = """The proxy schematic does not contain the correct type.
+                         The provided type is {0}. The function requires a valid schematic as a argument
+                         This allows a response object to be tested against the schematic""" \
+                .format(str(type(self._response_model.response_schematic_validation)))
+            self.log.error(message)
+            raise TypeError(message)
+
+    def _set_status_code_validation(self, status_code):
```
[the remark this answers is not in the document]
MelcomVE: Forgot about them they basically do nothing as it private and not used removed it

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+            return True
+
+        if isinstance(self._response_model.response_status_code_validation, int):
+            if response.status_code != self._response_model.response_status_code_validation:
+                message = """The response was unable to conform to the validation status code. The status code
+                             applied was {0} The expected status code is {1}
+                             To fix this you can either modify the status code validation or remove it entirely
+                             """.format(str(response.status_code),
+                                        str(self._response_model.response_status_code_validation))
+                self.log.error(message)
```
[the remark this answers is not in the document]
MelcomVE: I check it and I do log before almost every exception. But will remove if not required

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+    """
+    _request_model = HTTPHelperRequestModel()
+    _authentication_types = [item.value for item in HTTPAuthenticationType]
+    _request_types = [item.value for item in HTTPRequestType]
+    _http_method_enums = [item.value for item in HTTPMethod]
+
+    def __init__(self, request_model=None):
+        self.log = logging.getLogger('%s.request_handler' % __name__)
+        # Receive a model or create one if not specified
+        self._request_model = HTTPHelperRequestModel(request_model) if request_model is not None \
```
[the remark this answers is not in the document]
MelcomVE: It has checking when constructed but that is not an full check. Running validate does an extensive check about the content inside the dict and even if there are required items missing

stackstate_checks_base/tests/test_http_helper.py
MelcomVE: Replaced with pytest

stackstate_checks_base/tests/test_http_helper.py
MelcomVE: Replaced with pytest

stackstate_checks_base/tests/test_http_helper.py
MelcomVE: I followed the structure for most of the other integration libs, will replace

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1222 @@
...
+            @session_handler
+                The Session Handler Class. This allows the user to also pass down a custom session handler if required
+            @request_handler
+                The Request Handler Class. This allows the user to also pass down a custom request handler if required
+        """
+
+        session = session_handler.create_session_object()
+        request = request_handler.create_request_object()
+
+        # Mock requests
```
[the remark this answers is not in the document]
MelcomVE: Will note in the future

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: Added a response Connection response handler which will be the last point of exposure for the user. get_session is added to that

hruhek (changes requested):

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1005 @@
...
+from schematics.types import BaseType, StringType, IntType, DictType, BooleanType
+from schematics.exceptions import DataError
+from requests import Session, Request
+from requests.auth import HTTPBasicAuth
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+from ..utils.common import ensure_string, ensure_unicode, to_string
+import sys
+
+PY2 = sys.version_info[0] == 2
```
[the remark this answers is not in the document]
MelcomVE: Removed, This is deprecated with the Unicode fix

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1005 @@
...
+        self._response_model = HTTPHelperResponseModel(response_model) if response_model is not None \
+            else HTTPHelperResponseModel()
+
+    def get_status_code(self):
+        """
+        Return the status code response from the executed Session() object
+        """
+        return self._response_model.response.status_code
+
+    def get_json(self):
```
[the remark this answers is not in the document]
MelcomVE: That has been removed with the Unicode fix

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,477 @@
...
+        self.assertEqual(result.get_request_headers(), {'a': '1', 'Content-Length': '100'})
+
+
+class TestHTTPHelperRequestHandler(unittest.TestCase):
+    def test_get_request_model(self):
+        request_handler = HTTPHelperRequestHandler()
+        model = request_handler.get_request_model()
+        self.assertTrue(isinstance(model, HTTPHelperRequestModel))
+
+        with pytest.raises(DataError):
```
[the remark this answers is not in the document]
MelcomVE: pytest has been swapped out

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,1005 @@
...
+        else:
+            message = """The headers provided does not contain the correct type.
+                         The provided type is {0}. This function only accepts dict
+                         objects this allows a easy mapping to the query object as the headers
+                         also exists out of key and value pairs. The current headers passed
+                         was the following {1}""".format(str(type(self._request_model.headers)),
+                                                         str(self._request_model.headers))
+            raise TypeError(message)
+
+    def validate_auth(self):
```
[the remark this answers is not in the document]
MelcomVE: Added a standalone function

hruhek (approved):

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,929 @@
...
+        This is the model for the `HTTPHelper` class.
+        When using the `HTTPHelper` class there will be the only acceptable values to create state and kick off the
+        request
+    """
+    mock_enable = BooleanType(required=False, default=False)
+    mock_status = IntType(required=False, default=None)
+    mock_text = BaseType(required=False, default=None)
+    endpoint = StringType(required=True, default=None)
+    auth_data = DictType(BaseType, required=False, default=None)
+    auth_type = BaseType(required=False, default=None)
```
lowdewijk, line 59: can be string with choices.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,929 @@
...
+            except TypeError as e:
+                message = """The request was unable to conform to the validation schematic. The
+                             schematic applied was {0}
+                             The error provided by the schematic validation is {1}
+                             To fix this you can either modify the schematic validation or remove it entirely
+                             """.format(str(self._request_model.request_schematic_validation), e)
+                raise TypeError(message)
+
+    @staticmethod
+    def split_string_into_dict(target, delimiter, sub_delimiter):
```
lowdewijk, line 419: is this still used somewhere?

stackstate_checks_base/tests/test_http_helper.py
```diff
@@ -0,0 +1,389 @@
...
+    hello = StringType(required=True)
+    pong = BooleanType(required=True)
+
+
+class TestHTTPHelperBase(unittest.TestCase):
+    """
+     HTTP Helper Request Class
+    """
+
+    def test_compact_unicode_response(self):
```
lowdewijk, line 37: This test only proofs that you can deal with unicode characters in UTF-8, but this does not proof you can also deal with different unicode encodings.

lowdewijk (commented):
In general I am seeing that you've not really utilized schematics to do the actual validation for you. You still have a lot of validate methods that are just type checking, when schematics has all kinds of types for checking complexer types. 

Also I don't see that you're consistently calling validate on the models you have. Why have the models when you don't call validate on them? I believe you can get rid of a lot of code if you would just let schematics do its work. Also, there is no reason to delay it. When you create the model or pass the model you can immediately validate it. The longer you wait with the validation the more chance that you forget to do it at all (which seems to be the case also).

Lastly, I am fairly sure unicode is taken care of properly by requests, but I am still missing a test that shows that when a _different_ encoding than UTF-8 comes over the wire, that our HttpHelper is able to handle it. If a HTTP header contains for example: `Content-Type:text/html; charset=ISO-8859-1`. Does it then properly decode the string if some weird `ISO-8859-1` characters are used?

To make matters worse there is also the weird way in which Python 2 stores unicode strings. Fortunately we can switch to Python 3 with the Agentv3 soon.

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,929 @@
...
+                self._response_model.response_schematic_validation(parsed_json) \
+                    .validate()
+
+                return True
+
+            except DataError as e:
+                message = """The response was unable to conform to the validation schematic.
+                             The error provided by the schematic validation is {0}
+                             To fix this you can either modify the schematic validation or remove it entirely
+                             """.format(e)
```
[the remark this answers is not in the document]
MelcomVE: Better safe than sorry, Wrapped them in str

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,929 @@
...
+        else:
+            message = """The headers provided does not contain the correct type.
+                         The provided type is {0}. This function only accepts dict
+                         objects this allows a easy mapping to the query object as the headers
+                         also exists out of key and value pairs. The current headers passed
+                         was the following {1}""".format(str(type(self._request_model.headers)),
+                                                         str(self._request_model.headers))
+            raise TypeError(message)
+
+    def validate_auth(self):
```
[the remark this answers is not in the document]
MelcomVE: Yes, that is true forgot to include it back in the create objects. Added it into the execution cycle

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
MelcomVE: It appears not. I'm removing this forcing the user of the library to send the correct information in the correct fields

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,929 @@
...
+        When using the `HTTPHelperRequestHandler` class there will be the only acceptable values to create state, This
+        state can in turn be used within the HTTPHelper class
+    """
+    method = StringType(required=True)
+    endpoint = StringType(required=True, default=None)
+    query = DictType(StringType, default=dict({}))
+    headers = DictType(StringType, default=dict({}))
+    body = BaseType(required=False, default=None)
+    auth_data = DictType(BaseType, required=False, default=None)
+    auth_type = BaseType(required=False, default=None)
```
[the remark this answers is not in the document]
MelcomVE: I converted both enums that was strings only to rather use the stringType in the schematic

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,929 @@
...
+            - SET && GET HTTP Timeout
+            - SET && GET HTTP Retry Policy
+            - SET && GET HTTP SSL Verification
+            - SET && GET HTTP Proxy
+            - Sending the HTTP Request and Session to the Endpoint
+    """
+    log = logging.getLogger('%s.connection_handler' % __name__)
+
+    def __init__(self, connection_model=None):
+        self._connection_model = HTTPHelperConnectionModel(connection_model) if connection_model is not None \
```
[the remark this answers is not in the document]
MelcomVE: I added it into the original _builder function now called _build_validate_and_execute
This function creates the handlers, validates them and then executes the request

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,929 @@
...
+            return True
+
+        elif isinstance(self._connection_model.retry_policy, dict):
+            Retry(**self._connection_model.retry_policy)
+            return True
+
+        return False
+
+    def validate_ssl_verify(self):
+        if isinstance(self._connection_model.ssl_verify, bool):
```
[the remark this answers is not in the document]
MelcomVE: True I removed validate_timeout and validate_ssl_verify
They tested for an int and bool

stackstate_checks_base/stackstate_checks/base/utils/http_helper.py
```diff
@@ -0,0 +1,786 @@
...
+        # This will work on straight text or json objects
+        # This should take the object from a unicode state to a dict state
+        json_decoded_response = json.loads(ensured_string_values)
+
+        # If valid json was found return this json
+        # Run validation before we return the value
+        if issubclass(type(json_decoded_response), dict) and self.validate_body_schematic(json_decoded_response):
+            return json_decoded_response
+        else:
+            return None
```
lowdewijk, line 671: This is a silent failure. Under what circumstances is this a correct answer? Raise an exception.

MelcomVE: I usually prefer the Some None approach instead of forcing errors if possible

PR #123, opened 2021-03-25
Stac 12220 - make state location configurable from check config

zandrewitte:
Step 1: Link to Jira issue
https://stackstate.atlassian.net/browse/STAC-12220

Step 2: Description of changes

    Make the check state location configurable in the conf.d of the check.
    Make Check State and Auto Snapshotting Agent Check mixins.

- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


stackstate_checks_base/stackstate_checks/base/checks/base.py
```diff
@@ -905,6 +906,15 @@ def _is_not_empty(self, field):
 
         return False
 
+    def get_check_state_path(self):
+        """
+        get_check_state_path returns the path where the check state is stored. By default the check configuration
+        location will be used. If state_location is set in the check configuration that will be used instead.
+        """
+        state_location = self.instance.get("state_location", self.get_agent_conf_d_path())
```
vishalxebia, line 914: @zandrewitte What will happen if the path doesn't exist or problem with the permission of writing? Is it already taken care?

vishalxebia, line 914: @zandrewitte What will happen if the path doesn't exist or problem with the permission of writing? Is it already taken care?

vishalxebia (changes requested):

stackstate_checks_base/stackstate_checks/base/checks/base.py
zandrewitte: The read part is fine actually, it was when it was trying to clear the state on a read-only file system that it failed. So I added a check to see if the file exists before trying to remove it. If there is a file and it fails to remove it, it should actually result in an error

vishalxebia (approved):

PR #132, opened 2021-04-22
Stac 12217 multi aws account region

Eelviny:
https://stackstate.atlassian.net/browse/STAC-12217

Implement new IAM role assuming functionality, where the agent uses one global IAM access key/secret access key, then assumes a role into a each instance.

Provide a region list in each instance, and iterate through each region to fetch topology for every region in the list. A special "global" region exists and the registry categorizes global services. Global services are not queried in normal regions.

There are some concerns about handling large amounts of API calls and the limits of Python for parallelizing different regions, this will be tackled soon. For now each region is run in serial.

- [x] Unit/Component test
- [x] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/aws_topology/.gitignore
+++ b/aws_topology/.gitignore
@@ -1,0 +1,1 @@
+run_aws_topology.py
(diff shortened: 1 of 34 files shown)


Jacco (approved):
Checked: Does not touch any thing other then aws_toplogy

PR #145, opened 2021-05-25
Stac 13058 component rename

Eelviny:
### Step 1: Link to Jira issue
https://stackstate.atlassian.net/browse/STAC-13058

### Step 2: Description of changes
Rename the SQS, SNS and Kinesis components to have more friendly user-facing names for each component, instead of using IDs, ARNs or other.

### Step 3: Did you add / update tests for your changes in the right area?
- [x] Unit/Component test
- [ ] Integration test	

### Step 4: I'm confident that everything is properly tested:
I got a PO / QA Approval by:
- [ ] Name

### Step 5: Can we ship this feature to production?
- [ ] Yes, I'm proud of my work. Ship it! :ship:

PR #147, opened 2021-05-27
Stac 13125 aws ci

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13125

Add aws_topology to the gitlab-ci.yaml, and fix some cross-region/cross-account errors for SQS and API Gateway from #145 that got merged into master unnoticed as the pipeline didn't run the tests.

- [x] Unit/Component test
- [x] Integration test	

I got a PO / QA Approval by:
- [ ] Name

- [x] Yes, I'm proud of my work. Ship it! :ship:

### DIFF
--- a/aws_topology/stackstate_checks/aws_topology/resources/api_gateway.py
+++ b/aws_topology/stackstate_checks/aws_topology/resources/api_gateway.py
@@ -133,13 +133,11 @@ class ApigatewayStageCollector(RegisteredResourceCollector):
                                     method_integration_uri.rfind('arn'):method_integration_uri.find('/invocations')
                                 ]
                             elif re.match("arn:aws:apigateway:.+:sqs:path/.+", method_integration_uri):
-                                #  TODO cross region fails
-                                queue_name = method_integration_uri.rsplit('/', 1)[-1]
                                 queue_arn = arn(
-                                    'sqs',
-                                    self.location_info['Location']['AwsRegion'],
-                                    self.location_info['Location']['AwsAccount'],
-                                    queue_name
+                                    resource='sqs',
+                                    region=method_integration_uri.split(':', 4)[3],
+                                    account_id=method_integration_uri.split('/', 2)[1],
+                                    resource_id=method_integration_uri.rsplit('/', 1)[-1],
                                 )
                                 integration_arn = queue_arn
 
--- a/aws_topology/stackstate_checks/aws_topology/resources/sqs.py
+++ b/aws_topology/stackstate_checks/aws_topology/resources/sqs.py
@@ -2,15 +2,20 @@ from .utils import make_valid_data, with_dimensions, CloudTrailEventBase, create
 from .registry import RegisteredResourceCollector
 from schematics import Model
 from schematics.types import StringType, ModelType
+import re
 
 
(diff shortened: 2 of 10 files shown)


aws_topology/stackstate_checks/aws_topology/resources/api_gateway.py
```diff
@@ -133,13 +133,11 @@ def process_all(self, filter=None):
...
                             elif re.match("arn:aws:apigateway:.+:sqs:path/.+", method_integration_uri):
-                                #  TODO cross region fails
-                                queue_name = method_integration_uri.rsplit('/', 1)[-1]
                                 queue_arn = arn(
-                                    'sqs',
-                                    self.location_info['Location']['AwsRegion'],
-                                    self.location_info['Location']['AwsAccount'],
-                                    queue_name
+                                    resource='sqs',
+                                    region=method_integration_uri.split(':', 4)[3],
```
zandrewitte, line 138: You have access to the `location_info` right? That looks a lot safer than splitting on the uri. If the location is not an option, just check these splittings before looking up the index they might fail pretty easily.

aws_topology/stackstate_checks/aws_topology/resources/sqs.py
```diff
@@ -4,16 +4,12 @@
...
-    return url.rsplit('/', 1)[-1]
-
-
 def create_arn(region=None, account_id=None, resource_id=None, **kwargs):
     return arn(
         resource='sqs',
-        region=region,
-        account_id=account_id,
-        resource_id=get_queue_name_from_url(resource_id),
+        region=resource_id.split('.', 2)[1],
```
zandrewitte, line 10: Why are we ignoring the `region`, `account_id` that is being passed in? Do the splitting before calling this function if they are indeed different

aws_topology/tests/test_template.py
```diff
@@ -863,20 +863,20 @@ def test_process_kinesis_streams(self):
         self.assertEqual(len(topology), 1)
         self.assertEqual(len(topology[0]["relations"]), 0)
         self.assertEqual(len(topology[0]["components"]), 4)
-        self.assertEqual(topology[0]["components"][0]["Name"], "stream_1")
+        self.assertEqual(topology[0]["components"][0]["data"]["Name"], "stream_1")
```
zandrewitte, line 866: Was the left side just wrong and you're fixing it? If not and we're changing the location of values like name we should make sure the StackPack expects them there

zandrewitte (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/api_gateway.py
Eelviny: This is very much on purpose, because if we use `location_info` we assume that the target resource is in the same AWS account and region. This breaks when the resource exists in a different account or region

aws_topology/stackstate_checks/aws_topology/resources/sqs.py
Eelviny: Jacco and I had a discussion and we need to rework how we pass in these values, as using them breaks when the resource exists within another account and region. For now this resolves the issue

zandrewitte (approved):

PR #157, opened 2021-06-04
Refactor ELB Classic, use ARN for identifier

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13201

- Refactor ELB classic
- Implement IAM permission graceful degredation
- Improve API call efficiency by fetching ELBs and tags in batches
- Use ARN of ELB for primary identifier
- Fix incorrect tests
- Fix autoscaling identifiers for ARN

- [x] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


aws_topology/tests/json/template/test_elbv2_describe_tags.json
```diff
@@ -1,7 +1,7 @@
 {
     "TagDescriptions": [
         {
-            "ResourceArn": "arn:foo:bar",
+            "LoadBalancerName": "classic-loadbalancer-1",
```
Eelviny, line 4: I made an incorrect change here - the real problem is that we assumed that elb and elbv2 had the same API response. This is not the case. I will revert this and instead make a new template file

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -13,78 +42,105 @@ class ELBClassicCollector(RegisteredResourceCollector):
...
+            **kwargs
+        ):
+            yield elb_page.get("LoadBalancerDescriptions", [])
+
+    @set_required_access_v2("elasticloadbalancing:DescribeInstanceHealth")
+    def collect_instance_health(self, elb_name):
+        return self.client.describe_instance_health(LoadBalancerName=elb_name).get("InstanceStates", [])
+
+    def process_elbs(self, elb_names=[]):
+        if elb_names:  # Only pass in LoadBalancerNames if a specific name is needed, otherwise ask for all
```
Jacco, line 68: I would rather have an optional page processor callback passed into client_array_operation

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -13,78 +42,105 @@ class ELBClassicCollector(RegisteredResourceCollector):
...
+        elb_arn = self.agent.create_arn(
+            "AWS::ElasticLoadBalancing::LoadBalancer", self.location_info, resource_id=elb_name
+        )
+        output["Name"] = elb_name
+        output["Tags"] = data.tags
+        output["URN"] = [elb_arn]
+        self.emit_component(elb_arn, self.COMPONENT_TYPE, output)
+
+        vpc_id = elb.VPCId
+        self.emit_relation(elb_arn, vpc_id, "uses service", {})
```
Jacco, line 106: Should we change id?

Jacco (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
Eelviny: We use ARNs everywhere else, I found the formatting that was used before to be very strange

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -13,78 +42,105 @@ class ELBClassicCollector(RegisteredResourceCollector):
...
+
+    def process_one_elb(self, elb_name):
+        self.process_elbs([elb_name])
+
+    @transformation()
+    def process_elb(self, data):
+        elb = LoadBalancer(data.elb, strict=False)
+        elb.validate()
+        output = make_valid_data(data.elb)
+        elb_name = elb.LoadBalancerName
```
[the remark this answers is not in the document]
Eelviny: I meant to set the Name parameter of the resource, forgot to. Will use it there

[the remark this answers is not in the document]
Eelviny: Will change it around

aws_topology/stackstate_checks/aws_topology/resources/autoscaling.py
```diff
@@ -83,16 +83,19 @@ def process_autoscaling_group(self, data):
             self.emit_relation(auto_scaling_group.AutoScalingGroupARN, instance.InstanceId, 'uses service', {})
 
         for load_balancer_name in auto_scaling_group.LoadBalancerNames:
+            elb_arn = self.agent.create_arn(
+                "AWS::ElasticLoadBalancing::LoadBalancer", self.location_info, resource_id=load_balancer_name
```
zandrewitte, line 87: Is there not a way to re-use the loadbalancer arn function that is already defined?

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -13,78 +42,105 @@ class ELBClassicCollector(RegisteredResourceCollector):
     COMPONENT_TYPE = "aws.elb_classic"
     CLOUDFORMATION_TYPE = "AWS::ElasticLoadBalancing::LoadBalancer"
 
+    @set_required_access_v2("elasticloadbalancing:DescribeTags")
+    def collect_tag_page(self, elb_names):
+        if len(elb_names) <= 20:
```
zandrewitte, line 47: Either add a comment here about what this 20 means (api limit) or create a class variable so it's a bit more clear

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -13,78 +42,105 @@ class ELBClassicCollector(RegisteredResourceCollector):
...
+            return self.client.describe_tags(LoadBalancerNames=elb_names).get("TagDescriptions", [])
+        else:
+            return []
+
+    @set_required_access_v2("elasticloadbalancing:DescribeLoadBalancers")
+    def collect_elb_page(self, **kwargs):
+        # This doesn't use the utils.client_array_operation function as we need control over grabbing one page at a time
+        for elb_page in self.client.get_paginator("describe_load_balancers").paginate(
+            PaginationConfig={
+                "MaxItems": 20,  # Use 20 to match the describe_tags API call, where max items in one call is also 20
```
zandrewitte, line 57: I'm not sure whether we want to put a limitation on this as well. We can probably still fetch more here and batch them into 20 for the tag collection

aws_topology/stackstate_checks/aws_topology/resources/autoscaling.py
Eelviny: That's what this does, but it goes through the collector first. It does actually call the create_arn at the top of the file, but it's done this way because many times we need to do cross-API calls to create an ARN (for example when creating a relation) so we use the same method everywhere.

So the entire thing is abstracted into the collector. Normally we get the ARN back in the API call, but loadbalancing is an old API and is an exception

Jacco: I am planning to change how it all works in a later stage. The difference between using the agent and using the method directly is that we change change that around more easily. The agent will always know how to resolve, the implementaion might change in future.

Eelviny:
Blocked by #152

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -1,90 +1,151 @@
...
+
+class LoadBalancer(Model):
+    LoadBalancerName = StringType(required=True)
+    VPCId = StringType()
+    Instances = ListType(ModelType(LoadBalancerInstances))
+
+
+class InstanceHealth(Model):
+    InstanceId = StringType(required=True)
+    State = StringType()
```
Jacco, line 36: State can have default "UNKNOWN"

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -1,90 +1,151 @@
...
+    MAX_TAG_CALLS = 20  # This is the max items that can be requested in one describe_tags call
+
+    @set_required_access_v2("elasticloadbalancing:DescribeTags")
+    def collect_tag_page(self, elb_names):
+        max_items = self.MAX_TAG_CALLS  # Limit max items that we fetch to ensure call doesn't fail completely
+        return self.client.describe_tags(LoadBalancerNames=elb_names[:max_items]).get("TagDescriptions", [])
+
+    @set_required_access_v2("elasticloadbalancing:DescribeLoadBalancers")
+    def collect_elbs(self, **kwargs):
+        for elb in client_array_operation(self.client, "describe_load_balancers", "LoadBalancerDescriptions", **kwargs):
```
Jacco, line 54: Shouldn't the collect_instance_health be done in this loop?

Jacco (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
Eelviny: Does it need to be? I collect all resources in the same function together, process_elb_page

Jacco: Lets visually discuss. Can't see any more what I meant here :-)

zandrewitte (approved):

Jacco (approved):

PR #159, opened 2021-06-04
Refactor & Improve Route53

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13208

- Refactor route53_hostedzone and route53_domains
- Add create_arn functions for route53 and route53domains
- Fetch more domain information
- Add Schematics on API calls
- Add IAM permission checking
- Make route53 hosted zone name more user-friendly

- [x] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


Eelviny:
Blocked by #152

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
```diff
@@ -1,40 +1,104 @@
...
+    Id = StringType(required=True)
+    Name = StringType(default="UNKNOWN")
+
+
+class DelegationSet(Model):
+    Id = StringType()
+
+
+class VPCs(Model):
+    VPCId = StringType()
```
Jacco, line 28: Think this one should be required

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
```diff
@@ -1,40 +1,104 @@
...
+    return arn(resource="route53", region="", account_id="", resource_id="hostedzone/" + resource_id)
+
+
+class HostedZoneInfo(Model):
+    Id = StringType(required=True)
+    Name = StringType(default="UNKNOWN")
+
+
+class DelegationSet(Model):
+    Id = StringType()
```
Jacco, line 24: Required

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
```diff
@@ -1,40 +1,104 @@
...
 
 class Route53HostedzoneCollector(RegisteredResourceCollector):
     API = "route53"
     API_TYPE = "global"
     COMPONENT_TYPE = "aws.route53.hostedzone"
 
+    @set_required_access_v2("route53:ListTagsForResource")
+    def collect_tags(self, hosted_zone_id):
+        return (
+            self.client.list_tags_for_resource(ResourceType="hostedzone", ResourceId=hosted_zone_id)
```
Jacco, line 48: Tags collections often do support paginate. We need to check this for all APIs. Often they have a "HasMoreTags" attribute. Don't know if that supports the can_paginate call.

Jacco (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
Eelviny: This particular resource [does not](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/route53.html#Route53.Client.list_tags_for_resource) - let's worry about this with a pass over all resources later

Eelviny:
This is failing due to the tests needing to be fixed, but will wait until #152 is merged

aws_topology/stackstate_checks/aws_topology/resources/route53_domain.py
```diff
@@ -1,25 +1,77 @@
...
+    @transformation()
+    def process_domain(self, data):
+        domain = Domain(data.domain, strict=False)
+        domain.validate()
+        output = make_valid_data(data.domain)
+        domain_name = domain.DomainName
+        output["Name"] = domain_name
+        output["Tags"] = data.tags
+        output["URN"] = [
+            self.agent.create_arn("AWS::Route53Domains::Domain", self.location_info, resource_id=domain_name)
```
Jacco, line 75: Here we do not use the CLOUDFORMATION_TYPE (we want to get rid of)

aws_topology/stackstate_checks/aws_topology/resources/route53_domain.py
```diff
@@ -1,25 +1,77 @@
...
         ]
-        self.emit_component(domain_name, self.COMPONENT_TYPE, domain)
+        self.emit_component(domain_name, self.COMPONENT_TYPE, output)
```
Jacco, line 77: We need to make sure that a CF-template resources that is a route53 domain gets references correctly. (just verify route)

Jacco (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/route53_domain.py
Eelviny: Actually route53 domains is not supported by Cfn... I kinda made up the Cfn type, according to how it would be

Jacco (approved):

aws_topology/stackstate_checks/aws_topology/resources/route53_domain.py
```diff
@@ -1,25 +1,77 @@
...
+    def collect_tags(self, domain_name):
+        return self.client.list_tags_for_domain(DomainName=domain_name).get("TagList", [])
+
+    @set_required_access_v2("route53domains:GetDomainDetail")
+    def collect_domain_description(self, domain_name):
+        return self.client.get_domain_detail(DomainName=domain_name) or {}
+
+    def collect_domain(self, domain_data):
+        domain_name = domain_data.get("DomainName", "")
+        data = deepcopy(domain_data)
```
zandrewitte, line 41: I think the `deepcopy` is unnecessary, we're not sharing this dict right?

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
```diff
@@ -1,40 +1,103 @@
...
-            "arn:aws:route53:::{}".format(hosted_zone_id.lstrip('/'))
+        if not filter or "hostedzones" in filter:
+            self.process_hosted_zones()
+
+    @transformation()
+    def process_hosted_zone(self, data):
+        hosted_zone = HostedZone(data.hosted_zone, strict=False)
+        hosted_zone.validate()
+        output = make_valid_data(data.hosted_zone)
+        hosted_zone_id = hosted_zone.HostedZone.Id.rsplit("/", 1)[-1]
```
zandrewitte, line 96: Can this potentially break?

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
```diff
@@ -1,40 +1,103 @@
...
+from schematics.types import StringType, ModelType, ListType
+
+
+def create_arn(region=None, account_id=None, resource_id=None, **kwargs):
+    return arn(resource="route53", region="", account_id="", resource_id="hostedzone/" + resource_id)
+
+
+class HostedZoneInfo(Model):
+    Id = StringType(required=True)
+    Name = StringType(default="UNKNOWN")
```
zandrewitte, line 20: Do we really need this default? If we can't get the name can we uniquely identify this hosted zone?

zandrewitte: I see we use the hosted zone id as the external id, maybe we can fallback to that instead of unknown for the name

zandrewitte: You will probably have to keep this here then and add a check in `process_hosted_zone`

zandrewitte (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
Eelviny: Unlikely - if a value is not found, Schematics has a default value of `UNKNOWN`. If no forward slash is found, then it will return the whole string as usual

aws_topology/stackstate_checks/aws_topology/resources/route53_domain.py
Eelviny: In general I've been taught that you should not mutate a reference that is passed into a function, but it won't break anything if I don't do it.

I don't mind either way, let me know and I'll remove it

aws_topology/stackstate_checks/aws_topology/resources/route53_hostedzone.py
Eelviny: The reason for this is the format is `/hostedzone/<id of resource>` - this just removes the first section

zandrewitte (approved):

PR #161, opened 2021-06-07
Refactor RDS, add tags to cluster

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13222

- Refactor RDS
- Add IAM permission checks
- Add Schematics
- Collect tags for Clusters too

- [x] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


Eelviny:
Waiting on #152 before adding/fixing tests

aws_topology/stackstate_checks/aws_topology/resources/rds.py
```diff
@@ -1,89 +1,161 @@
...
+    @transformation()
+    def process_instance(self, data):
+        instance = Instance(data.instance, strict=False)
+        instance.validate()
+        output = make_valid_data(data.instance)
+        instance_arn = instance.DBInstanceArn
+        instance_id = instance.DBInstanceIdentifier
+        output["Name"] = instance_id
+        output["Tags"] = data.tags
+        output["URN"] = ["urn:endpoint:/" + instance.Endpoint.Address]
```
Jacco, line 123: If you dereference without any check then Endpoint Model should have required=True on the InstanceModel (or have a default)

That way any of the exists checks will already be checked on the "validate" step

aws_topology/stackstate_checks/aws_topology/resources/rds.py
```diff
@@ -1,89 +1,161 @@
...
+        output.update(with_dimensions([{"key": "DBInstanceIdentifier", "value": instance_id}]))
+        self.emit_component(instance_arn, "aws.rds_instance", output)
+        self.emit_relation(instance_arn, instance.DBSubnetGroup.VpcId, "uses service", {})
+        # TODO agent.create_security_group_relations (but needs change?)
+        for security_group in output.get("VpcSecurityGroups", []):
```
Jacco, line 128: Here the loop should not be ont the output. But rather on the schematic instance

aws_topology/stackstate_checks/aws_topology/resources/rds.py
```diff
@@ -1,89 +1,161 @@
...
+            self.process_instance(instance_data)
+
+    def process_all(self, filter=None):
+        if not filter or "clusters" in filter:
+            self.process_clusters()
+        if not filter or "instances" in filter:
+            self.process_instances()
+
+    def process_one_cluster(self, cluster_arn):
+        cluster_id = cluster_arn.rsplit(":", 1)[-1]
```
Jacco, line 105: Guess this can become a "safe" function in utils that returns "" in case the is no colon. (Also add test for that utils)

aws_topology/stackstate_checks/aws_topology/resources/rds.py
```diff
@@ -1,89 +1,161 @@
...
-        cluster_data['Name'] = cluster_arn
-        cluster_data.update(with_dimensions([{'key': 'DBClusterIdentifier', 'value': cluster_id}]))
-        self.emit_component(cluster_arn, self.COMPONENT_TYPE, cluster_data)
-        for cluster_member in cluster_data['DBClusterMembers']:
-            db_identifier = cluster_member['DBInstanceIdentifier']
-            arn = self.agent.create_arn('AWS::RDS::DBInstance', self.location_info, db_identifier)
-            self.emit_relation(cluster_arn, arn, 'has_cluster_node', {})
-
-    EVENT_SOURCE = 'rds.amazonaws.com'
+    @transformation()
```
Jacco, line 114: Note to self: Transormation currently does not ignore the exception (yet). Maybe we should add the ignore parameter as with @set_required_access_v2 and start ignoring only the schematics validation exception.

aws_topology/stackstate_checks/aws_topology/resources/rds.py
```diff
@@ -1,89 +1,161 @@
...
+class InstanceVpcSecurityGroup(Model):
+    VpcSecurityGroupId = StringType(required=True)
+
+
+class Instance(Model):
+    DBInstanceArn = StringType(required=True)
+    DBInstanceIdentifier = StringType(default="UNKNOWN")
+    DBSubnetGroup = ModelType(InstanceSubnetGroup)
+    Endpoint = ModelType(InstanceEndpoint)
+    VpcSecurityGroups = ListType(ModelType(InstanceVpcSecurityGroup))
```
Jacco, line 54: You could add a default of [] to VpcSecurityGroups

Jacco (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/rds.py
Eelviny: I've already made it required

aws_topology/stackstate_checks/aws_topology/resources/rds.py
Eelviny: I did a test and found that the API accepts either the identifier or the full ARN, so now I just pass in the full ARN

aws_topology/stackstate_checks/aws_topology/resources/rds.py
Eelviny: Let's make a note for a pass ;) currently I'm working on consistency and previous PRs don't have this

Jacco (approved):

zandrewitte (approved):

PR #163, opened 2021-06-09
S3 add IAM checks & fix STAC-13234

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13234
https://stackstate.atlassian.net/browse/STAC-13236

- Only add S3 bucket components where the bucket is in the same region as the check (previously all regions were returned in every check)
- Add IAM permissions checks to S3
- Return empty tag array if S3 returns error code `NoSuchTagSet`
- Expand test capabilities for arbitrary `ClientError`s
- Add test API response for `NoSuchTagSet`
- Remove old `ignore=True` permission value in set_required_access_v2, rework tests

- [x] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


aws_topology/stackstate_checks/aws_topology/resources/s3.py
```diff
@@ -55,30 +68,35 @@ def collect_buckets(self):
         ]:
             yield bucket
 
-    def process_all(self, filter=None):
+    @set_required_access_v2("s3:ListAllMyBuckets")
+    def process_buckets(self):
         # TODO buckets should only be fetched for global OR filtered by LocationConstraint
```
Jacco, line 73: The comment can go now :-)

aws_topology/stackstate_checks/aws_topology/resources/s3.py
```diff
@@ -27,23 +27,36 @@ class S3Collector(RegisteredResourceCollector):
     COMPONENT_TYPE = "aws.s3_bucket"
     CLOUDFORMATION_TYPE = 'AWS::S3::Bucket'
 
+    @set_required_access_v2("s3:GetBucketLocation")
```
Jacco, line 30: You need a different access right then I suppose

aws_topology/stackstate_checks/aws_topology/resources/s3.py
```diff
@@ -55,30 +68,35 @@ def collect_buckets(self):
         ]:
             yield bucket
 
-    def process_all(self, filter=None):
+    @set_required_access_v2("s3:ListAllMyBuckets")
+    def process_buckets(self):
         # TODO buckets should only be fetched for global OR filtered by LocationConstraint
-        if not filter or 'buckets' in filter:
```
Jacco, line 60: I think we still need the filter because there will be a lot more from this API later? Or do you want to add it later?

aws_topology/stackstate_checks/aws_topology/resources/utils.py
```diff
@@ -199,7 +199,7 @@ def is_throttling_error(code):
     return code in _THROTTLED_ERROR_CODES
 
 
-def set_required_access_v2(value, ignore=False):
+def set_required_access_v2(value, ignore=False, ignore_codes=[]):
```
Jacco, line 202: The naming is now confusing

Jacco (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/s3.py
Eelviny: This line is still in there, just got moved to another place - check the diff

aws_topology/stackstate_checks/aws_topology/resources/utils.py
Eelviny: What do you suggest instead?

vishalxebia (approved):

PR #170, opened 2021-06-11
Refactor and simplify cloudformation

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13273

- Refactor CloudFormation
- Add Schematics
- Add IAM permission checks
- Simplify logic - remove list_stacks command
- Create relation to child stacks via CloudFormation resources, not ParentId
- Bugfix: don't emit a relation for resources we don't support

- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


aws_topology/stackstate_checks/aws_topology/resources/cloudformation.py
```diff
@@ -65,41 +73,66 @@ def no_arn(region=None, account_id=None, resource_id=None, **kwargs):
...
+        self.emit_component(stack.StackId, self.COMPONENT_TYPE, output)
 
-    def process_resource(self, stack_id, resource):
-        resource_type = resource["ResourceType"]
-        if resource.get("PhysicalResourceId"):
-            arn = self.agent.create_arn(resource_type, self.location_info, resource.get("PhysicalResourceId"))
-            self.emit_relation(stack_id, arn, "has resource", {})
+        resources = [StackResource(resource, strict=False) for resource in data.resources]
+        for resource in resources:
+            resource.validate()
```
zandrewitte, line 133: If one `resource` fails, we will stop processing the rest. Is this the desired outcome here?

Eelviny: Not "desired" per se, but if we come across an API response we don't expect, it would probably not be a good idea to continue. We also exit out the same way on a per-resource basis, at the top of this function

Jacco:
This one needs to be fixed!

Eelviny:
Pipeline is failed because I manually stopped it, aws_topology check passed but I need pipeline resources elsewhere ;)

zandrewitte (approved):

PR #173, opened 2021-06-14
Send aws_topology snapshot before raising errors

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13287

Ensure that `stop_snapshot` is called before raising all found errors, to ensure that one failing API does not block the sending of successful APIs.

- [x] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/aws_topology/stackstate_checks/aws_topology/aws_topology.py
+++ b/aws_topology/stackstate_checks/aws_topology/aws_topology.py
@@ -142,14 +142,14 @@ class AwsTopologyCheck(AgentCheck):
                     }
                     self.event(event)
                     errors.append("API %s ended with exception: %s %s" % (spec["api"], str(e), traceback.format_exc()))
-            # TODO this should be for tests, in production these relations should not be sent out
-            agent_proxy.finalize_account_topology()
-            self.components_seen = agent_proxy.components_seen
-            if len(errors) > 0:
-                raise Exception("get_topology gave following exceptions: %s" % ", ".join(errors))
-        self.delete_ids += agent_proxy.delete_ids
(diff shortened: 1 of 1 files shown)


aws_topology/stackstate_checks/aws_topology/aws_topology.py
```diff
@@ -142,14 +142,14 @@ def get_topology(self, instance_info, aws_client):
...
-                raise Exception("get_topology gave following exceptions: %s" % ", ".join(errors))
-        self.delete_ids += agent_proxy.delete_ids
+        # TODO this should be for tests, in production these relations should not be sent out
+        agent_proxy.finalize_account_topology()
+        self.components_seen = agent_proxy.components_seen
 
         self.stop_snapshot()
+        if len(errors) > 0:
+            raise Exception("get_topology gave following exceptions: %s" % ", ".join(errors))
+        self.delete_ids += agent_proxy.delete_ids
```
Jacco, line 152: Move this to under components_seen :-)

Jacco (changes requested):

Eelviny:
Continuing from the discussion in Slack -  @zandrewitte you had a concern that we're attempting to close a snapshot even if there are errors.

Yes, we are attempting to close if there are errors, but only the errors that we specifically catch (like the accessdenied errors). Any uncaught errors will make the snapshot collection completely fail as normal.

Jacco:
Above statement is not how it is intended

Eelviny:
This actually doesn't fix the underlying issue, closing

PR #185, opened 2021-06-23
Make state non disk persisted

Jacco:

https://stackstate.atlassian.net/browse/STAC-13394

Make last_full_topology not persisted on disk

- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


aws_topology/stackstate_checks/aws_topology/aws_topology.py
```diff
@@ -65,8 +62,19 @@ def get_account_id(self, instance_info):
...
+            # Create empty state
+            self.log.info('  Result => YES (first run)')
+            return True
+        secs = seconds_ago(self.last_full_topology)
+        self.log.info('  Result => {} (Last run was {} seconds ago, interval is set to {})'.format(
+            "YES" if secs > interval else "NO",
+            int(secs),
+            interval)
+        )
+        return secs > interval
```
Eelviny, line 77: Pieter reported that the topology could be running every 61 minutes instead of 60 minutes, this could be the issue as it's expecting it to be more than the interval

aws_topology/stackstate_checks/aws_topology/aws_topology.py
```diff
@@ -161,17 +181,18 @@ def get_topology(self, instance_info, aws_client):
...
-            if len(errors) > 0:
-                raise Exception("get_topology gave following exceptions: %s" % ", ".join(errors))
+        # TODO this should be for tests, in production these relations should not be sent out
+        self.log.info('Finalize FULL scan (#components = {})'.format(len(agent_proxy.components_seen)))
+        agent_proxy.finalize_account_topology()
+        self.components_seen = agent_proxy.components_seen
         self.delete_ids += agent_proxy.delete_ids
 
         self.stop_snapshot()
+        if len(errors) > 0:
```
Eelviny, line 191: I can't remember if we decided to actually make this change where the errors are emitted after the snapshot - does it make any difference?

Eelviny (changes requested):

aws_topology/stackstate_checks/aws_topology/aws_topology.py
Jacco: I think this fixes this

aws_topology/stackstate_checks/aws_topology/aws_topology.py
Jacco: I think this makes more sense

Eelviny (approved):

aws_topology/stackstate_checks/aws_topology/aws_topology.py
```diff
@@ -82,28 +90,30 @@ def check(self, instance_info):
...
+                self.service_check(self.SERVICE_CHECK_EXECUTE_NAME, AgentCheck.OK, tags=instance_info.tags)
+            except Exception as e:
+                msg = "AWS topology collection failed: {}".format(e)
+                self.log.error(msg)
+                self.service_check(
+                    self.SERVICE_CHECK_EXECUTE_NAME, AgentCheck.WARNING, message=msg, tags=instance_info.tags
+                )
+
+        try:
             self.get_topology_update(instance_info, aws_client)
```
zandrewitte, line 110: Do we always run the update straight after the full scan? We can just wait for the next check run right

aws_topology/stackstate_checks/aws_topology/aws_topology.py
```diff
@@ -161,17 +181,18 @@ def get_topology(self, instance_info, aws_client):
                     }
                     self.event(event)
                     errors.append("API %s ended with exception: %s %s" % (spec["api"], str(e), traceback.format_exc()))
-            # TODO this should be for tests, in production these relations should not be sent out
-            agent_proxy.finalize_account_topology()
-            self.components_seen = agent_proxy.components_seen
-            if len(errors) > 0:
-                raise Exception("get_topology gave following exceptions: %s" % ", ".join(errors))
+        # TODO this should be for tests, in production these relations should not be sent out
+        self.log.info('Finalize FULL scan (#components = {})'.format(len(agent_proxy.components_seen)))
```
zandrewitte, line 185: Will this print all of the components that was seen? In that case I would make it debug log

aws_topology/stackstate_checks/aws_topology/aws_topology.py
zandrewitte: Broken record :smile: You really shouldn't do this. It will cause flickering topology. Let's figure out why we're currently seeing the flickering, fix it and then keep the pattern that is used across all integrations

aws_topology/stackstate_checks/aws_topology/aws_topology.py
Jacco: Only the number of components of this scan will be shown

zandrewitte (approved):

PR #186, opened 2021-06-24
Refactor + Improve EC2

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13384
https://stackstate.atlassian.net/browse/STAC-13385
https://stackstate.atlassian.net/browse/STAC-13386

- Refactor EC2
- Use Schematics for fetching values
- Add IAM permission checks
- Fix pagination issues with fetching resources
- Fix issue where bare metal instances were considered as Nitro hypervisor instances
- Fetch instance types on demand instead of fetching the entire list, then hold in memory
- Use API result filtering to reduce data retrieved
- Process the full EC2 instance when receiving a state-change event instead of just sending an agent event
- Add new update events: StartInstances, StopInstances, TerminateInstances, ModifyInstanceAttribute, AttachVolume, DetachVolume
- Only create a relation between a VPN Gateway and a VPC if the State is attached
- Only create VPN gateway components if the VPN gateway is pending or available, not deleting or deleted

- [x] Unit/Component test
- [x] Integration test	

I got a PO / QA Approval by:
- [ ] Name


aws_topology/stackstate_checks/aws_topology/resources/ec2.py
```diff
@@ -207,48 +311,58 @@ def process_subnet(self, subnet_description):
...
+    def process_vpn_gateway(self, data):
+        vpn_gateway = VpnGateway(data, strict=False)
+        vpn_gateway.validate()
+        output = make_valid_data(data)
+        self.emit_component(vpn_gateway.VpnGatewayId, "aws.vpngateway", output)
+        for vpn_attachment in vpn_gateway.VpcAttachments:
+            if vpn_attachment.State == "attached":
+                self.emit_relation(vpn_gateway.VpnGatewayId, vpn_attachment.VpcId, "uses service", {})
+
+    def process_batch_instances(self, event, seen):
```
Jacco, line 337: Because we do a schema validation it needs the @transformation to ignore it

aws_topology/stackstate_checks/aws_topology/resources/ec2.py
```diff
@@ -207,48 +311,58 @@ def process_subnet(self, subnet_description):
...
-            for vpn_attachment in vpn_description["VpcAttachments"]:
-                vpc_id = vpn_attachment["VpcId"]
-                self.emit_relation(vpn_id, vpc_id, "uses service", {})
-
-    def process_run_instances(self, event, seen):
-        ids = []
+        for vpn_gateway_data in self.collect_vpn_gateways():
+            self.process_vpn_gateway(vpn_gateway_data)
+
+    def process_vpn_gateway(self, data):
```
Jacco, line 328: @transformation

Jacco (approved):

aws_topology/stackstate_checks/aws_topology/resources/ec2.py
```diff
@@ -53,119 +103,163 @@ def process_all(self, filter=None):
...
         if not filter or "vpcs" in filter:
             self.process_vpcs()
+        if not filter or "subnets" in filter:
+            self.process_subnets()
         if not filter or "vpn_gateways" in filter:
             self.process_vpn_gateways()
 
+    @set_required_access_v2("ec2:DescribeInstanceTypes")
+    def collect_instance_types(self):
+        # This list does not change often, safe to hold in memory
```
zandrewitte, line 113: Looks like it indeed make sense to keep this in memory, but I think we should have an expiration for this then? Currently the only way this is emptied is by restarting the agent if I understand this correctly?

Eelviny: Good point - I realise that the list can grow, but the items in it never change. I changed this around so I only grab an instance type one at a time, when I need it, then add it to a dict. If it's already in the dict we read it from the dict, otherwise we go and get it from the API. It means that we can always grab new instance types, and speeds up processing as I'm not grabbing the entire list.

zandrewitte (approved):

PR #194, opened 2021-07-01
Fix ELB reduced permission cases

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13502

- Fix issue where an unhandled API error would occur if permission `elasticloadbalancing:DescribeLoadBalancers` is removed
- Fix issue where a NoneType error would occur if permission `elasticloadbalancing:DescribeTags` is removed

- [x] Unit/Component test
- [x] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
+++ b/aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
@@ -49,7 +49,6 @@ class ELBClassicCollector(RegisteredResourceCollector):
         max_items = self.MAX_TAG_CALLS  # Limit max items that we fetch to ensure call doesn't fail completely
         return self.client.describe_tags(LoadBalancerNames=elb_names[:max_items]).get("TagDescriptions", [])
 
-    @set_required_access_v2("elasticloadbalancing:DescribeLoadBalancers")
     def collect_elbs(self, **kwargs):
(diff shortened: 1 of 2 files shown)


aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
```diff
@@ -69,6 +68,7 @@ def process_elb_page(self, elbs):
                     tags = tag_result.get("Tags", [])
             self.process_elb(LoadBalancerData(elb=data, tags=tags, instance_health=instance_health))
 
+    @set_required_access_v2("elasticloadbalancing:DescribeLoadBalancers")
```
vishalxebia, line 71: @Eelviny  should we put this decorator on `collect_elbs` function which is actually calling `DescribeLoadBalancers` API?

vishalxebia (changes requested):

aws_topology/stackstate_checks/aws_topology/resources/elb_classic.py
Eelviny: Not here - this change is in line with all other resources. collect_elbs is a paginator function to abstract the describe_load_balancers call, so we need to place the set_required_access on the function that calls the paginator

vishalxebia (approved):

PR #195, opened 2021-07-01
Fix DynamoDB tag and description collection with reduced perms

Eelviny:
https://stackstate.atlassian.net/browse/STAC-13511

- Fix DynamoDB description construction when `dynamodb:DescribeTable` is missing
- Fix tag collection
- Add tag collection API response tests

- [x] Unit/Component test
- [x] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
+++ b/aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
@@ -56,23 +56,16 @@ class DynamodbTableCollector(RegisteredResourceCollector):
 
     def construct_table_description(self, table_name):
         return {
-            "Table": {
-                "TableName": table_name,
-                "TableArn": self.agent.create_arn("AWS::DynamoDB::Table", self.location_info, resource_id=table_name),
-            }
+            "TableName": table_name,
+            "TableArn": self.agent.create_arn("AWS::DynamoDB::Table", self.location_info, resource_id=table_name),
         }
 
     def collect_table(self, table_name):
         # If data collection fails, construct the ARN
         data = self.collect_table_description(table_name) or self.construct_table_description(table_name)
-        # If ARN creation fails, skip tag collection
-        if data.get("Table", {}).get("TableArn"):
-            tags = self.collect_tags(data["Table"]["TableArn"])
-        else:
-            tags = []
+        tags = self.collect_tags(data.get("TableArn")) or []
         return TableData(table=data, tags=tags)
 
-    @set_required_access_v2("dynamodb:ListTables")
     def collect_tables(self):
         for table in [
             self.collect_table(table_name)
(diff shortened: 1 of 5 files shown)


aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
```diff
@@ -56,37 +56,31 @@ def collect_table_description(self, table_name):
...
 
-    @set_required_access_v2("dynamodb:ListTables")
     def collect_tables(self):
         for table in [
             self.collect_table(table_name)
             for table_name in client_array_operation(self.client, "list_tables", "TableNames")
         ]:
             yield table
 
+    @set_required_access_v2("dynamodb:ListTables")
```
vishalxebia, line 76: this access is more valid on `collect_tables` function.

aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
```diff
@@ -56,37 +56,31 @@ def collect_table_description(self, table_name):
...
+    def process_tables(self):
+        for table_data in self.collect_tables():
+            self.process_table(table_data)
```
vishalxebia, line 79: Will this not throw an exception anymore since we were catching earlier?

aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
```diff
@@ -56,37 +56,31 @@ def collect_table_description(self, table_name):
 
     def construct_table_description(self, table_name):
         return {
-            "Table": {
-                "TableName": table_name,
-                "TableArn": self.agent.create_arn("AWS::DynamoDB::Table", self.location_info, resource_id=table_name),
-            }
+            "TableName": table_name,
+            "TableArn": self.agent.create_arn("AWS::DynamoDB::Table", self.location_info, resource_id=table_name),
```
vishalxebia, line 60: Are we trying to create the pre-assumed Table ARN since it already failed in the API call from boto? Why we want to get the false Table from our side?

aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
Eelviny: It breaks there. Check all other resources, this is intentional. If the collect_tables function fails, we need to stop the entire service collection

aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
Eelviny: This fixes the exception. It was not being caught in the right way, and was causing the service to fail out completely instead of just stopping the table execution and emitting a warning. The exception handling is in @set_required_access_v2

aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
Eelviny: This is for graceful degredation. If the describe_table_description call fails because of no permission, we still need the table ARN to be able to create relations. So, if we can't get the ARN from the actual API, we'll construct it instead.

vishalxebia: @Eelviny But it's ok if we can't create relations at least we know we couldn't retrieve an item due to a permission issue. We shouldn't ignore the actual problem may be.

aws_topology/stackstate_checks/aws_topology/resources/dynamodb.py
vishalxebia: @Eelviny Then I saw this somewhere that decorator was wrongly placed on the fucntion.

vishalxebia (approved):

PR #228, opened 2021-09-06
Stac 12434 raw agent metrics

MelcomVE:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/stackstate_checks_base/stackstate_checks/base/checks/base.py
+++ b/stackstate_checks_base/stackstate_checks/base/checks/base.py
@@ -857,6 +857,12 @@ class AgentCheckBase(object):
     def _submit_metric(self, mtype, name, value, tags=None, hostname=None, device_name=None):
         pass
 
+    def _submit_raw_metric(self, name, value, tags=None, hostname=None, device_name=None, timestamp=None):
+        pass
+
+    def raw(self, name, value, tags=None, hostname=None, device_name=None, timestamp=None):
+        self._submit_raw_metric(name, value, tags, hostname, device_name, timestamp)
+
     def gauge(self, name, value, tags=None, hostname=None, device_name=None):
         self._submit_metric(aggregator.GAUGE, name, value, tags=tags, hostname=hostname, device_name=device_name)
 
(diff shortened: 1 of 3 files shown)


agent_integration_sample/stackstate_checks/agent_integration_sample/agent_integration_sample.py
```diff
@@ -192,3 +192,9 @@ def check(self, instance):
         self.health.start_snapshot()
         self.health.check_state("id", "name", Health.CRITICAL, "identifier", "msg")
         self.health.stop_snapshot()
+
+        # raw metrics
```
zandrewitte, line 196: Add one sending in a `hostname` as well

stackstate_checks_base/stackstate_checks/base/checks/base.py
```diff
@@ -1106,6 +1112,19 @@ def _submit_metric(self, mtype, name, value, tags=None, hostname=None, device_na
 
         aggregator.submit_metric(self, self.check_id, mtype, ensure_unicode(name), value, tags, hostname)
 
+    def _submit_raw_metrics_data(self, name, value, tags=None, hostname=None, device_name=None, timestamp=None):
+        tags = self._normalize_tags_type(tags, device_name, name)
+
+        if timestamp is None:
+            self.warning('Raw Metric: Timestamp is invalid')
```
zandrewitte, line 1119: `Raw Metric: Timestamp is invalid`

`Timestamp is empty, not sending metric: x` something like that. Just to make it clear it's not being sent and it's empty rather than invalid

stackstate_checks_base/stackstate_checks/base/checks/base.py
```diff
@@ -1106,6 +1112,19 @@ def _submit_metric(self, mtype, name, value, tags=None, hostname=None, device_na
...
+            return
+
+        if hostname is None:
+            hostname = ''
+
+        self.log.debug('Raw Metric: Submitting Raw Metric Data')
```
zandrewitte, line 1125: This will be done by the batcher, so we can remove this here

PR #261, opened 2022-04-01
Trying to fix failing win build

hruhek:
Lazy-object-proxy in version 1.7.0 dropped python2 support.

https://pypi.org/project/lazy-object-proxy/


julianosk (approved):
LGTM

PR #285, opened 2022-11-25
Fix some splunk_base issues

YMvanderHorst:

No Jira issue, but here is StackState support request number #1189 
not sure how to directly link that

- Updated splunk_base so it actually looks into the selected app when making a list of saved searches
- Updated splunk_base so the return type for dispatched saved search is actually JSON instead of adding a useless parameter to the body of the dispatch request.

- [ ] Unit/Component test
- [ ] Integration test	
- [X] I checked but it doesn't affect any tests.

I got a PO / QA Approval by:
- [ ] Name

- [X] Yes, I'm proud of my work. Ship it! :ship:

### DIFF
--- a/splunk_base/stackstate_checks/splunk/client/splunk_client.py
+++ b/splunk_base/stackstate_checks/splunk/client/splunk_client.py
@@ -157,12 +157,13 @@ class SplunkClient:
         """ This method is mocked for testing. Do not change its behavior """
         return datetime.datetime.utcnow()
 
-    def saved_searches(self):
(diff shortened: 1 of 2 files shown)


LouisParkin:
Hi @YMvanderHorst 
The changes you made in this PR could not be merged without adjustments.  I cherry-picked your changes into our code, and it has been tested, merged, and released. The new tag for the docker images is 3.0.1 
Also, it is no longer the Python 2 agent, but Python 3, so the image name loses the `-2` and is now just stackstate-agent, cluster agent etc.

PR #296, opened 2023-11-01
STAC-20332 Updated some dependencies at random in order to generate a…

LouisParkin:
… base diff for an MR.

- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


.setup-scripts/image/Makefile
```diff
@@ -1,12 +1,27 @@
 RUNTIMETAG := $(shell date +%Y%m%d)
 
+#build:
```
craffit, line 3: some comments here, can we ditch these?

craffit (approved):

.setup-scripts/image/Makefile
LouisParkin: Doing.

PR #304, opened 2024-05-15
STAC-21039 Moving towards DataDog 7.51.1 tag.

LouisParkin:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


stackstate_checks_base/stackstate_checks/base/checks/prometheus/prometheus_base.py
```diff
@@ -23,7 +23,7 @@
 
 
 class PrometheusCheck(PrometheusScraperMixin, AgentCheck):
-    DEFAULT_METRIC_LIMIT = 2000
+    DEFAULT_METRIC_LIMIT = 25000
```
rb3ckers, line 26: Should we do a quick check on how much more data this sends in some cases (like for StackState itself)? Or was this already the setting before?

craffit: This has been in since may

PR #315, opened 2025-02-27
Stac 21521 stackstate 7.51.1

LouisParkin:
- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name


dynatrace_base/stackstate_checks/dynatrace/__init__.py
```diff
@@ -5,5 +5,5 @@
 
 __all__ = [
     '__version__',
-    'dynatrance_client'
+    'dynatrace_client.py'
```
craffit: nice catch:)

dynatrace_base/stackstate_checks/dynatrace/entity_types.py
```diff
@@ -0,0 +1,31 @@
+
```
craffit: nice putting this as a separate file

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
```diff
@@ -8,40 +8,26 @@
...
 
-DYNATRACE_UI_URLS = {
-    "SERVICE": "%s/#newservices/serviceOverview;id=%s",
-    "PROCESS_GROUP": "%s/#processgroupdetails;id=%s",
-    "PROCESS_GROUP_INSTANCE": "%s/#processdetails;id=%s",
-    "PROCESS": "%s/#processdetails;id=%s",
-    "HOST": "%s/#newhosts/hostdetails;id=%s",
-    "APPLICATION": "%s/#uemapplications/uemappmetrics;uemapplicationId=%s",
-    "CUSTOM_DEVICE": "%s/#customdevicegroupdetails/entity;id=%s"
+TOPOLOGY_API_SPEC = {
```
craffit: i am suprised to find TOPOLOGY_API_SPEC here, maybe call it EVENTS_API_SPEC or HEALTH_API_SPEC?

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
```diff
@@ -186,15 +188,18 @@ def _collect_events(self, dynatrace_client, instance_info):
         event_limit_reached = None
         try:
             while events_response:
-                events = events_response.get('events', [])
+                if type(events_response) is dict:
+                    events = events_response.get('events', [])
+                else:
+                    events = events_response
```
craffit, line 238: this is pretty much fail-silently, if the event_respons eis not what we expect. i'd rather throw/crash out here to not silently discard unexpected responses.

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
```diff
@@ -203,21 +208,21 @@ def _collect_events(self, dynatrace_client, instance_info):
...
         params = {}
         if from_time:
             params['from'] = from_time
-        if cursor:
-            params['cursor'] = cursor
-        endpoint = dynatrace_client.get_endpoint(url, "/api/v1/events")
+        if next_page_key:
+            params['nextPageKey'] = next_page_key
+        endpoint = dynatrace_client.get_endpoint(url, "/api/v2/events")
         events = dynatrace_client.get_dynatrace_json_response(endpoint, params)
```
craffit, line 279: how are errors communicated here? does this throw on non-200 responses such that we fail fast/hard?

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
```diff
@@ -245,10 +265,37 @@ def _collect_topology(self, response, component_type, instance_info):
...
+                    else:
+                        dynatrace_component = Entity.model_validate(item)
+                else:
+                    dynatrace_component = DynatraceComponent.model_validate(item)
+            except Exception as e:
+                self.log.warn("Couldn't create topology component: %s" % e)
+                print(f"{item}")
+                raise e
+
+            # try:
```
craffit: possible spurious comments

dynatrace_health/tests/test_dynatrace_health.py
```diff
@@ -3,166 +3,292 @@
...
-        assert expected_event == telemetry._topology_events[0]
+    assert len(aggregator.events) == 0
 
 
 @freeze_time('2021-02-16 14:26:24')
-def test_state_data(state_manager, dynatrace_check, test_instance, requests_mock, aggregator):
+def test_raise_exception_for_response_code_not_200(dynatrace_check, test_instance, requests_mock, aggregator):
     """
-    Check is the right timestamp is writen to the state
+    Test to raise a check exception when API endpoint when status code is not 200
```
craffit, line 285: ah this answered my earlier question

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
craffit: in what scenario was this case necessary?

scripts/dynatrace_ingestion/ingest.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+if [ "$#" -ne 3 ]; then
```
craffit, line 3: a small readme around this with what it is for, how to setup and how to excersice it would be great. Also, was this used to generate any of the reference data? If so it makes sense to describe how that reference was created through this.

craffit: i now see the additions to the readme, maybe expand a bit on that.

dynatrace_health/tests/samples/custom_alert_response.json
```diff
@@ -0,0 +1,159 @@
+{
```
craffit, line 1: i think this is the non-ai generated response it might make sense to document how this was created (aswell as the ai generated ones) to have our future selves understand where this came from and maybe reproduce

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
LouisParkin: I don't think it is actually used.  That was my initial thoughts around replicating the process I followed with the topology api, I'll confirm and remove it.

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
LouisParkin: I can't remember, this was an issue with the topology api, it's more pre-emptive than an actual tested case.  The thing is, it did happen once or twice, and adding this if-statement resolved it.  I don't see it as failing silently, since the code executes normally after this.  Think of it as hardening.

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
LouisParkin: I think the line you should be concerned about is line 226... and non-200 responses are dealt with inside `get_dynatrace_json_response` as follows:
```
                if response.status_code != 200:
                    if "error" in response_json:
                        msg = response_json["error"].get("message")
                    else:
                        msg = "Got %s when hitting %s" % (response.status_code, endpoint)
                    self.log.error(msg)
                    raise Exception(
                        'Got an unexpected error with status code %s and message: %s' % (response.status_code, msg))
```

dynatrace_topology/stackstate_checks/dynatrace_topology/dynatrace_topology.py
LouisParkin: The try block (which used to be one, valid for a single model, was moved into the `for loop` with nested `if-else` statements above it, which deals with multiple distinct model types. I will remove the commented code.

scripts/dynatrace_ingestion/ingest.sh
LouisParkin: @dottorblaster I believe this was your contribution.  Oblige the man :-)

dynatrace_health/stackstate_checks/dynatrace_health/dynatrace_health.py
LouisParkin: Was indeed unused. I removed it.

scripts/dynatrace_ingestion/ingest.sh
dottorblaster: @craffit thanks for the tip, I pushed a change to the README 👍

dottorblaster (approved):
LGTM! No blockers from my side 👍

dottorblaster (approved):
LGTM again

dynatrace_health/testing.md
```diff
@@ -0,0 +1,66 @@
+# Dynatrace Health Integration Test Data Generation
```
craffit, line 1: this explanation is missing how ingest.sh was used in generating the test data.

the reading here is also very fluffy, probably because it was ai generated.  probably super great to generate, not so great to review.

dynatrace_base/tests/test_dynatrace_client_factory.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.fixture
+def client_args():
+    return {
+        "instance_name": "https://instance.live.dynatrace.com",
+        "token": "original_token"
+    }
+
+
+def test_jwt_auth_enabled(client_args, requests_mock: Mocker):
```
craffit, line 19: greta testing on this level.

craffit (approved):

dynatrace_health/testing.md
LouisParkin: ingest.sh wasn't used to generate the data.  The AI looked at example JSON files for the v2 API already present in the repo, and then it went online to read Dynatrace's docs and specs, and then it proceeded to generate the required files without issue (and the test JSON files were correct and usable right off the bat).

PR #316, opened 2025-03-06
Update stepfunction.py

Wolgo:
Explicitly only retrieve METADATA_ONLY

Not setting this value can result in errors if the agent does not have access to the KMS key or the kms:decrypt action, while the description is always discarded.

Instead, we can prevent these errors and remove the need to remove the description.

See: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/stepfunctions/client/describe_state_machine.html

Only retrieve METADATA to prevent errors on permissions that are needed to retrieve data that are discarded later on. 

- [ ] Unit/Component test
- [ ] Integration test	

I got a PO / QA Approval by:
- [ ] Name

### DIFF
--- a/aws_topology/stackstate_checks/aws_topology/resources/stepfunction.py
+++ b/aws_topology/stackstate_checks/aws_topology/resources/stepfunction.py
@@ -75,7 +75,7 @@ class StepFunctionCollector(RegisteredResourceCollector):
 
     @set_required_access_v2("states:DescribeStateMachine")
     def collect_state_machine_description(self, arn):
-        return self.client.describe_state_machine(stateMachineArn=arn)
+        return self.client.describe_state_machine(stateMachineArn=arn, includedData='METADATA_ONLY')
 
     def collect_state_machine(self, summary):
         arn = summary.get("stateMachineArn")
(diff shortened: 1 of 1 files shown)


LouisParkin:
Thanks for this, and apologies for the long silence — it was not neglect alone. `master` required a status check (`ci/gitlab/gitlab.com`) from a GitLab pipeline that has since been archived, so this PR was literally unmergeable no matter who reviewed it. That is now fixed, and it needs only a review.

Tracked as [STAC-25522](https://stackstate.atlassian.net/browse/STAC-25522) (the protection fix) and [STAC-25580](https://stackstate.atlassian.net/browse/STAC-25580) (this change).

The reasoning holds up: the definition is fetched and then discarded, so `includedData='METADATA_ONLY'` removes both the wasted payload and the `kms:decrypt` requirement. Two things a reviewer will want to confirm — that our pinned botocore is new enough for `includedData`, and that the change gets forward-ported to `stackstate-7.78.2`, which is the branch that actually runs CI.

Wolgo:
Hey @LouisParkin, as far as I see the whole stepfunction and AWS_topology has been removed in 7.78.2. 
Sounds like you can just safely close the ticket (and this PR).

PR #326, opened 2025-08-12
STAC-23143: let the Splunk integration use multiple adapters for JWT auth

dottorblaster:
The Splunk integration could use JWT authentication, but only the one baked into Splunk itself. Since we want it to be portable to new authentication providers, I refactored the existing code into an adapter pattern.


LouisParkin (approved):
LGTM

PR #327, opened 2025-09-04
STAC-23310: rename MS_JWT_AUTH to SPLUNK_MS_JWT_AUTH in splunk client

dottorblaster:
Since we want some more fine grained control over the configuration of the JWT Auth both for dynatrace and the splunk client, we are going to rename the environment variable.


LouisParkin (approved):
LGTM

PR #362, opened 2026-08-03
STAC-25463 Add GitHub Actions check-test workflow (integrations CI migration phase 1)

LouisParkin:
Jira: https://stackstate.atlassian.net/browse/STAC-25463 (parent: [STAC-25142](https://stackstate.atlassian.net/browse/STAC-25142))

First phase of moving `stackstate-agent-integrations` CI off GitLab. **Purely additive** — `.gitlab-ci.yml` is untouched and keeps running, so this can land without a cutover.

## What moved

| GitLab | GitHub Actions |
|---|---|
| the `test_<check>` job family (20 near-identical jobs) | one `check-tests` matrix |
| per-job `changes:` rules | `.github/scripts/select-checks.sh` (plain `git diff` vs merge base) |
| `checksdev validate *` riding inside `test_stackstate_checks_base` | its own `validate` job |
| `services: docker:24-dind` + `DOCKER_HOST: tcp://docker:2375` | dropped — both public ARC runners already ship a DinD sidecar |

Path filtering is done in `git` rather than a path-filter action because StackVista enforces a strict third-party action allowlist, and `git diff` against the merge base is precisely what the GitLab rule meant.

## Scoped out (each reasoned in the workflow header)

- **The 5 Docker-dependent suites** — `splunk_{base,health,metric,topology}` + `stackstate_checks_dev`. These are the only suites needing a Docker daemon. Phase 2; the runners already provide DinD, so it is about exercising that path, not provisioning it.
- **`test_postgres` — dead config.** It tests a `postgres` check that does not exist in this repository. Dropped, not pending.
- **`print_env` — not ported at any phase.** It is a bare `printenv`, and this repo is **public**, so it publishes every CI credential in scope to a world-readable log. Same hazard class as the `sign_debian_package.sh` finding on the agent.
- **No Windows lane exists to port.** `.gitlab-ci.yml` defines a `.windows_env` anchor that **no job has ever referenced**, and Windows is not a supported target. The orphaned `conda_env.ps1` / `windows_*.cmd` helpers can retire with the GitLab pipeline (STAC-25464).
- **Publishing + the runner-image `docker` job** stay on GitLab until integrations gets its own publishing role (pulumi-infra section 7.4).
- **Cerberus notify** — unlike the agent, this pipeline has never had a notify job, so adding one is new capability, and `CERBERUS_LAMBDA_URL` is a private-visibility org secret unreadable from a public repo. Follow-up.

## Credentials

Phase 1 needs **no new secret provisioning**: `REGISTRY_HOST` / `REGISTRY_USER` / `REGISTRY_PASSWORD` are org-level `visibility: all` (STAC-25350) and already reach this public repo. Note the image ref composes `vars.REGISTRY_HOST` + `/docker` rather than using `vars.REGISTRY_DOCKER_URL`, which is private-visibility and **not** readable here.

## Validation

- `actionlint` clean — added `.github/actionlint.yaml` declaring the self-hosted `*-public` labels
- `zizmor` clean (2 suppressed = the documented `unpinned-images` ignores on the container ref)
- `shellcheck` clean
- Selector exercised against real git history:

  | scenario | selected |
  |---|---|
  | `push` / dispatch | all 15 |
  | PR touching only `zabbix/` | `["zabbix"]` |
  | PR touching `kubelet/` + `vsphere/` | `["kubelet","vsphere"]` |
  | PR touching a shared library | all 15 |
  | PR touching nothing relevant | `[]` (matrix skips) |

**This PR self-validates** — the workflow runs against this very PR, so the image path and runner labels are proven before merge.

### DIFF
--- a/.github/workflows/checks-tests.yml
+++ b/.github/workflows/checks-tests.yml
@@ -19,7 +19,10 @@ name: Check tests
 #       sidecar, so this is about exercising that path rather than provisioning
 #       anything -- phase 2. That phase also brings across
 #       .setup-scripts/setup_artifactory_docker.sh (the registry docker login)
-#       and COMPOSE_HTTP_TIMEOUT, which only those suites need.
+#       and COMPOSE_HTTP_TIMEOUT, which only those suites need. It will also need
+#       a docker client, which the BCI Python image used here does not carry --
+#       either a BCI image with docker added, or a DinD service container, rather
+#       than a return to the private python:3.13.14-bookworm runner image.
 #   test_postgres
 #       Dead config: .gitlab-ci.yml tests a `postgres` check that does not exist
 #       in this repository. Dropped, not pending.
@@ -44,21 +47,29 @@ name: Check tests
 #       PUBLIC repo without a pulumi-infra grant. Tracked as a follow-up.
 #
 # CREDENTIALS
-#   Pulling the runner image needs vars.REGISTRY_HOST / vars.REGISTRY_USER and
-#   secrets.REGISTRY_PASSWORD, which are org-level visibility=all (STAC-25350)
-#   and therefore already reach this PUBLIC repo. Note it composes
-#   vars.REGISTRY_HOST + /docker rather than using vars.REGISTRY_DOCKER_URL: that
-#   variable is private-visibility and so is NOT readable here. Org secrets are
-#   never exposed to fork PRs, hence the same-repo guard on every job that pulls
-#   the image.
+#   The container image is SUSE BCI from registry.suse.com, which is public, so
+#   these jobs need no registry credentials at all. That is deliberate: this is a
+#   PUBLIC repository, and every job here executes PR-authored workflow, setup and
+#   test code. Any secret exposed to that code is exposed to whoever can open a
+#   branch. The earlier design pulled a private runner image with
+#   vars.REGISTRY_USER / secrets.REGISTRY_PASSWORD; dropping it removes the
+#   registry password from the PR path entirely and, as a side effect, lets
+#   Dependabot PRs run -- they receive no Actions secrets, so the image pull
+#   could never have succeeded for them.
 #
-#   Resolving check requirements additionally needs the private PyPI index, for
-#   the handful of pins that are not on public PyPI. That is
+#   One credential remains, and only on the step that needs it: the read-only
+#   pull from the private PyPI index, for pins that public PyPI does not serve
+#   (currently vsphere-automation-sdk). That is
 #   vars.GITLAB_PACKAGE_REGISTRY_PYPI_SIMPLE_URL and
 #   secrets.GITLAB_PACKAGE_REGISTRY_USER, granted to this repo in pulumi-infra
 #   (StackVista/pulumi-infra#263), alongside the already-org-wide
-#   secrets.GITLAB_PACKAGE_REGISTRY_READONLY_PASSWORD. Pull-only: this repo's
-#   *publishing* role is still deferred, per the note above.
+#   secrets.GITLAB_PACKAGE_REGISTRY_READONLY_PASSWORD. It is pull-only and
+#   least-privilege by construction; this repo's *publishing* role is still
+#   deferred, per the note above. The step is guarded so its absence degrades to
+#   a warning rather than failing suites that do not use the index.
+#
+#   Fork PRs are still excluded from the credentialed jobs. Secret exposure is no
+#   longer the reason -- running unreviewed code on self-hosted runners is.
 
 on:
   pull_request:
@@ -68,6 +79,13 @@ on:
   push:
     branches:
       - stackstate-7.78.2
+    # GitLab's `release_branch` anchor (`if: $CI_COMMIT_TAG`) put every test job
+    # in .base_integration_rules on tag pipelines too, so releases carry the same
+    # verdict as the branch they cut from. Tags here are `<version>-<increment>`
+    # (7.78.2-4), but the rule was any-tag and this stays faithful to it: `**`
+    # matches tag names containing `/`, which a bare `*` would silently skip.
+    tags:
+      - '**'
   workflow_dispatch:
 
 permissions:
@@ -82,22 +100,40 @@ defaults:
     shell: bash
 
 env:
-  # Tag of the prebuilt test image. Kept in lockstep with RUNNER_IMAGE_TAG in
-  # .gitlab-ci.yml until the GitLab pipeline is retired; the image is still built
-  # and published by that pipeline's `docker` job.
-  RUNNER_IMAGE_TAG: 20260625-py313
+  # SUSE BCI Python, per the mandatory BCI base-image policy. Digest-pinned so a
+  # run is reproducible and so Zizmor's unpinned-images audit passes; the comment
+  # carries the human-readable version the digest resolves to.
+  #
+  # This deliberately replaces the GitLab pipeline's
+  # stackstate-agent-integrations-runner image (FROM python:3.13.14-bookworm).
+  # That image exists mainly to carry Docker CE and docker-compose, which only the
+  # phase-2 Splunk / checks_dev suites use; the phase-1 suites need a Python
+  # toolchain and nothing else. Dropping it also drops the registry credentials
+  # these jobs used to need -- see CREDENTIALS below.
+  #
+  # BCI publishes 3.13.13, one patch behind the 3.13.14 the agent embeds and the
+  # GitLab image pinned. CPython patch releases are bugfix-only, and the full
+  # phase-1 matrix (including vsphere against the private index) was verified
+  # green on 3.13.13 before this switch. Worth realigning when BCI ships .14.
+  BCI_PYTHON_IMAGE: registry.suse.com/bci/python:3.13@sha256:7d36dd3ba6596fb690e31d956952059fd010604ad6309f06462c02c4c9c01461 # 3.13.13
+
+  # Packages the BCI image does not ship but the toolchain build needs: cython and
+  # pyyaml==6.0.1 have no cp313 wheels and are compiled from source.
+  BCI_BUILD_PACKAGES: gcc python313-devel libffi-devel
 
 jobs:
   select-checks:
     name: Select check suites to run
-    # Not strictly credential-bound, but there is nothing to select for a fork PR
-    # whose downstream jobs cannot pull the image anyway.
+    # Fork PRs never run: this is a PUBLIC repository on self-hosted runners, and
+    # a fork branch is unreviewed code. The jobs no longer carry registry
+    # credentials, but running arbitrary PR code on our own runners is a separate
+    # concern from secret exposure, so the boundary stays.
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: docker-public
     timeout-minutes: 10
     outputs:
       checks: ${{ steps.select.outputs.checks }}
-      runner_image: ${{ steps.image.outputs.ref }}
+      image: ${{ steps.image.outputs.ref }}
     steps:
       - name: Check out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
@@ -107,15 +143,14 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
-      - name: Resolve the runner image reference
+      - name: Resolve the container image reference
         id: image
-        env:
-          REGISTRY_HOST: ${{ vars.REGISTRY_HOST }}
+        # `container.image` cannot read the `env` context, so the pin defined once
+        # at workflow level is republished here as an output the container jobs can
+        # reference. Keeps a single source of truth for the digest.
         run: |
           set -euo pipefail
-          # Same image the GitLab pipeline runs on, via the same read-only proxy:
-          # GitLab's ${REGISTRY_DOCKER_URL} is registry.tooling.stackstate.io/docker.
-          echo "ref=${REGISTRY_HOST}/docker/stackstate/stackstate-agent-integrations-runner:${RUNNER_IMAGE_TAG}" >>"${GITHUB_OUTPUT}"
+          echo "ref=${BCI_PYTHON_IMAGE}" >>"${GITHUB_OUTPUT}"
 
       - name: Select check suites
         id: select
@@ -139,18 +174,21 @@ jobs:
     runs-on: docker-public
     timeout-minutes: 30
     container:
-      # Tag-pinned rather than digest-pinned: the tag is produced by the GitLab
-      # `docker` job and updated in lockstep with .gitlab-ci.yml, so pinning a
-      # digest here would silently drift from the pipeline that builds it.
-      image: ${{ needs.select-checks.outputs.runner_image }} # zizmor: ignore[unpinned-images]
-      credentials:
-        username: ${{ vars.REGISTRY_USER }}
-        password: ${{ secrets.REGISTRY_PASSWORD }}
+      # Digest-pinned at workflow level (BCI_PYTHON_IMAGE); the ignore is only
+      # because Zizmor cannot follow the pin through a job output.
+      image: ${{ needs.select-checks.outputs.image }} # zizmor: ignore[unpinned-images]
     steps:
       - name: Check out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+
+      - name: Install toolchain build dependencies
+        run: |
+          set -eo pipefail
+          zypper --non-interactive --gpg-auto-import-keys refresh
+          # shellcheck disable=SC2086 # deliberately word-split into package args
+          zypper --non-interactive install ${BCI_BUILD_PACKAGES}
 
       - name: Build the toolchain virtualenv
         run: |
@@ -188,15 +226,21 @@ jobs:
       matrix:
         check: ${{ fromJson(needs.select-checks.outputs.checks) }}
     container:
-      image: ${{ needs.select-checks.outputs.runner_image }} # zizmor: ignore[unpinned-images]
-      credentials:
-        username: ${{ vars.REGISTRY_USER }}
-        password: ${{ secrets.REGISTRY_PASSWORD }}
+      # Digest-pinned at workflow level (BCI_PYTHON_IMAGE); the ignore is only
+      # because Zizmor cannot follow the pin through a job output.
+      image: ${{ needs.select-checks.outputs.image }} # zizmor: ignore[unpinned-images]
     steps:
       - name: Check out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
+
+      - name: Install toolchain build dependencies
+        run: |
+          set -eo pipefail
+          zypper --non-interactive --gpg-auto-import-keys refresh
+          # shellcheck disable=SC2086 # deliberately word-split into package args
+          zypper --non-interactive install ${BCI_BUILD_PACKAGES}
 
       - name: Build the toolchain virtualenv
         run: |
@@ -254,3 +298,87 @@ jobs:
           set -eo pipefail
           source venv/bin/activate
           checksdev test "${CHECK}" --bench
+
+  workflow-security:
+    name: Workflow security scan (Zizmor)
+    # Same fork boundary as the rest: this repository is PUBLIC and the runners are
+    # self-hosted, so unreviewed fork code is not checked out here even for a
+    # read-only audit. Depends on select-checks only to reuse the pinned image.
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    needs: select-checks
+    runs-on: docker-public
+    timeout-minutes: 15
+    container:
+      # Digest-pinned at workflow level (BCI_PYTHON_IMAGE); the ignore is only
+      # because Zizmor cannot follow the pin through a job output.
+      image: ${{ needs.select-checks.outputs.image }} # zizmor: ignore[unpinned-images]
+    permissions:
+      contents: read
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Install Zizmor
+        env:
+          ZIZMOR_VERSION: 1.28.0
+        run: |
+          set -eo pipefail
+          python3.13 -m venv /tmp/zizmor-venv
+          /tmp/zizmor-venv/bin/pip install --quiet "zizmor==${ZIZMOR_VERSION}"
+
+      - name: Zizmor audit
+        run: |
+          set -eo pipefail
+          # Offline: the online audits need a GitHub token, and handing this job a
+          # token to scan PR-authored workflows is the very shape Zizmor exists to
+          # catch. Persona `regular` keeps it to findings worth blocking on.
+          /tmp/zizmor-venv/bin/zizmor \
+            --persona regular \
+            --collect=workflows,actions \
+            .
+
+  ci-success:
+    name: CI success
+    # The single stable status for branch protection. Every other status here is
+    # either dynamically named (the matrix legs are `Check tests (<suite>)`) or
+    # conditional (check-tests is skipped entirely when the selector returns an
+    # empty array, and everything is skipped for fork PRs). A required check must
+    # exist on every run, so this job always runs and decides the verdict itself.
+    # That includes fork PRs, where everything upstream is skipped: without a
+    # status here a fork PR could never satisfy branch protection. It checks out
+    # nothing and runs no repository code, so it stays safe to run in that context.
+    if: always()
+    needs:
+      - select-checks
+      - validate
+      - check-tests
+      - workflow-security
+    runs-on: docker-public
+    timeout-minutes: 5
+    steps:
+      - name: Evaluate upstream job results
+        env:
+          # `skipped` is a legitimate outcome and must not be conflated with
+          # success: check-tests skips when no suite was selected, and the whole
+          # pipeline skips for fork PRs. Only `failure` and `cancelled` fail here.
+          RESULTS: >-
+            select-checks=${{ needs.select-checks.result }}
+            validate=${{ needs.validate.result }}
+            check-tests=${{ needs.check-tests.result }}
+            workflow-security=${{ needs.workflow-security.result }}
+        run: |
+          set -euo pipefail
+          failed=0
+          for entry in ${RESULTS}; do
+            echo "  ${entry}"
+            case "${entry#*=}" in
+              failure | cancelled) failed=1 ;;
+            esac
+          done
+          if [ "${failed}" -ne 0 ]; then
+            echo "::error title=CI failed::One or more required jobs did not succeed (see above)."
+            exit 1
+          fi
+          echo "All required jobs succeeded or were legitimately skipped."
--- a/.github/workflows/checks-tests.yml
+++ b/.github/workflows/checks-tests.yml
@@ -71,33 +71,46 @@ name: Check tests
 #   least-privilege by construction; this repo's *publishing* role is still
 #   deferred, per the note above.
 #
-#   Scoping it to a single step was not enough on its own, because the old setup
-#   script wrote it to ~/.netrc and the suite's own test code runs afterwards in
-#   the same job and could read it. So the suites that need it are split into
-#   their own job, `check-tests-private-index`, and that job now downloads the
-#   private packages into a local wheelhouse and deletes the credential before any
-#   repository code runs (STAC-25540). The remaining suites never see it.
+#   No pull request ever receives it. `pull_request` runs execute the pull
+#   request's own copy of this workflow and of every script it calls, so a run
+#   that holds a secret can always be made to disclose it -- by editing the
+#   fetching script, reordering steps, or adding one. A repository secret and
+#   pull-request-controlled code cannot be arranged into a boundary. The suites
+#   that need this credential therefore do not run on pull requests at all; they
+#   run on push, tag and workflow_dispatch events, whose contents are reviewed
+#   before reaching the release branch. See `check-tests-private-index`.
 #
-#   This replaced the `private-package-index` GitHub Environment, whose required
-#   reviewers released the credential only after a human approved each run. It
-#   worked, but SHARED_PATHS covers the CI files, so it fired on roughly one
-#   commit in six and blocked authors on their own pull requests for no gain the
-#   wheelhouse does not also provide.
+#   Within those runs the credential is still confined to a single step
+#   (STAC-25540): the script writes ~/.netrc, downloads one fixed package set into
+#   a local wheelhouse, deletes the netrc, and points pip at the wheelhouse, so
+#   the suite and its dependency tree install with nothing to authenticate
+#   against. The predecessor left the netrc readable for the rest of the job. That
+#   is defence in depth, not the boundary -- the boundary is the event condition
+#   above.
 #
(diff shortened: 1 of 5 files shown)


LouisParkin:
-

LouisParkin:
-

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,256 @@
...
+    strategy:
+      # One suite's failure should not mask the state of the others.
+      fail-fast: false
+      matrix:
+        check: ${{ fromJson(needs.select-checks.outputs.checks) }}
+    container:
+      image: ${{ needs.select-checks.outputs.runner_image }} # zizmor: ignore[unpinned-images]
+      credentials:
+        username: ${{ vars.REGISTRY_USER }}
+        password: ${{ secrets.REGISTRY_PASSWORD }}
```
LouisLotter: [P1] These tests execute PR-controlled workflow, setup, and test code while making registry and private-PyPI credentials available. An in-repo branch is still unreviewed, so the same-repository condition is not an approval boundary. Dependabot also passes this condition but receives no Actions secrets, causing the private image pull to fail. Please make PR tests credential-free, for example with a GitHub-hosted runner and BCI image, or approval-gate the credentialed path.

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,256 @@
...
+          checksdev validate manifest --include-extras
+          checksdev validate metadata
+          checksdev validate service-checks
+
+  check-tests:
+    name: Check tests (${{ matrix.check }})
+    if: >-
+      ${{ (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
+      && needs.select-checks.outputs.checks != '[]' }}
+    needs: select-checks
```
LouisLotter, line 268: [P2] The only test statuses are dynamically named matrix legs, and this job disappears when the selector returns an empty array. Branch protection cannot require these reliably without either waiting for absent suite names or leaving failures advisory. Please add an always-present ci-success job that evaluates select-checks, validate, check-tests, and a required Zizmor job.

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,256 @@
...
+#   *publishing* role is still deferred, per the note above.
+
+on:
+  pull_request:
+  # Mirrors GitLab's `master_branch` rule, which hardcoded the release branch the
+  # same way: the full matrix runs there regardless of what a given commit
+  # touched, so the branch always has a complete verdict.
+  push:
+    branches:
+      - stackstate-7.78.2
```
LouisLotter, line 126: [P2] The GitLab test jobs run for every CI_COMMIT_TAG, but defining only push.branches prevents this workflow from running for tag pushes. Please add an appropriate tags filter so release tags retain the migrated test coverage.

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,256 @@
...
+    # matrix, so it runs on every change rather than per suite.
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    needs: select-checks
+    runs-on: docker-public
+    timeout-minutes: 30
+    container:
+      # Tag-pinned rather than digest-pinned: the tag is produced by the GitLab
+      # `docker` job and updated in lockstep with .gitlab-ci.yml, so pinning a
+      # digest here would silently drift from the pipeline that builds it.
+      image: ${{ needs.select-checks.outputs.runner_image }} # zizmor: ignore[unpinned-images]
```
LouisLotter: [P2] The resolved runner image is built from python:3.13.14-bookworm. Introducing it as a job container conflicts with the mandatory SUSE BCI policy, and no no-BCI-alternative justification is documented. Phase-one suites need only Python and public PyPI, so please use a suitable BCI/shared BCI image or document a verified exception.

LouisLotter (changes requested):
Requesting changes for one credential-exposure issue and three CI migration gaps: a stable required gate, release-tag coverage, and BCI compliance.

LouisParkin:
-

.github/workflows/checks-tests.yml
LouisParkin: Fixed, and confirmed it was a real gap rather than a theoretical one: `.base_integration_rules` composes `master_branch` **and** `release_branch`, the latter being `if: $CI_COMMIT_TAG`, so every `test_<check>` job did run on tag pipelines.

Added a `tags` filter. Used `**` rather than `*` deliberately — GitHub`s `*` does not match `/`, and while tags here are `<version>-<increment>` (`7.78.2-4`), GitLab`s rule was any-tag and this stays faithful to it.

No selector change needed: it already treats every non-`pull_request` event as a full run, which is what the tag pipeline did.

.github/workflows/checks-tests.yml
LouisParkin: Fixed by switching to `registry.suse.com/bci/python`, digest-pinned.

Worth stating why this was clean rather than a fight: that runner image exists mainly to carry Docker CE and docker-compose, and those are only used by the phase-2 Splunk / checks_dev suites. Phase 1 needs a Python toolchain and nothing else, so there was no reason to inherit it.

Verified rather than assumed, before changing the workflow:

- BCI ships `/usr/bin/python3.13`, which `setup_env.sh` hardcodes
- it does **not** ship gcc, so `gcc python313-devel libffi-devel` are installed — `cython` and `pyyaml==6.0.1` have no cp313 wheels and compile from source
- ran the full phase-1 matrix in that image locally, including `vsphere` against the private index — all green

One honest caveat: **BCI tops out at 3.13.13, and the agent embeds 3.13.14**, which is exactly what the old image pinned. CPython patch releases are bugfix-only and the matrix is green, so I judged this acceptable, but it is a real difference and I have noted it in the workflow so it can be realigned when BCI ships `.14`. Say the word if you would rather not carry that gap.

Phase 2 will need a docker client that BCI Python does not have — noted in the header as "BCI + docker, or a DinD service", not as a reason to go back to the private image.

.github/workflows/checks-tests.yml
LouisParkin: Agreed, and fixed by removing the credentials rather than gating them.

The registry password was only ever needed to pull that private runner image. Moving to public BCI (see the base-image thread) means `vars.REGISTRY_USER` / `secrets.REGISTRY_PASSWORD` are gone from this workflow entirely — no approval gate to get wrong, nothing for PR-authored code to reach.

Your Dependabot point was correct and not hypothetical — this repo has active `Dependabot Updates` runs. Those PRs get no Actions secrets, so the image pull could never have succeeded. They work now because nothing needs a secret.

What remains is the read-only private **PyPI pull**, kept because `vsphere-automation-sdk` is not on public PyPI. It is scoped to the single step that needs it, credentials go to `~/.netrc` rather than the index URL, and it is guarded so an absent credential warns instead of failing unrelated suites. That is the least-privilege pattern rather than a broad grant, but it is still a secret on the PR path — flagging it plainly rather than claiming the surface is zero.

The same-repo guard stays. Its justification changed, though, and the header now says so: it is no longer about secrets but about not running unreviewed code on self-hosted runners.

.github/workflows/checks-tests.yml
LouisParkin: Added, both parts.

`ci-success` runs with `if: always()`, needs all four upstream jobs, and evaluates their results itself. It treats `skipped` as legitimate — `check-tests` skips when the selector returns `[]`, and the whole pipeline skips for fork PRs — while `failure` and `cancelled` fail the gate. That gives branch protection one stable, always-present status instead of dynamically named legs.

It deliberately checks out nothing and runs no repository code, which is what makes it safe on fork PRs. That matters more than it first appears: for a fork PR every other job is skipped, so without this the required check would never report at all and the PR could never go green.

Zizmor added as its own required job, running `--persona regular --collect=workflows,actions`. Kept offline on purpose — the online audits want a token, and handing a token to a job that scans PR-authored workflows is the exact shape Zizmor exists to catch.

Both green on the latest run, along with the full matrix.

LouisParkin:
-

LouisParkin:
### Follow-up found while this PR was waiting: branch protection here is broken independently of this change

While checking what actually gates merges on this repo, I found the branch protection is in a bad state, and it makes the `ci-success` job requested in review load-bearing rather than cosmetic. Filed as **STAC-25522** (blocked by this PR).

Verified today:

| Branch | Required checks | Effect |
| --- | --- | --- |
| `master` | `ci/gitlab/gitlab.com` | **Permanently unmergeable** |
| `legacy-agent-support` | none | CI gates nothing |
| `stackstate-7.51.1` | none | CI gates nothing |
| `stackstate-7.71.2` | none | CI gates nothing |
| `stackstate-7.78.2` (this PR's base) | none | CI gates nothing |

The GitLab project was archived, so nothing can ever report `ci/gitlab/gitlab.com` again. `master`'s HEAD commit status is `pending` with zero contexts, and PR #316 shows an empty check rollup while still sitting at `BLOCKED` — it can never merge. This is the same failure mode STAC-25420 already fixed for `stackstate-process-agent`.

The release branches have the opposite problem: they require nothing, so the workflow in this PR would run and be ignored by branch protection.

The fix (STAC-25522) is to adopt this repo into `pulumi-infra` — it is currently absent from `github/repositories/resources.yaml` entirely, so today's settings are unmanaged drift — and require the aggregate **`CI success`** context, mirroring the `stackstate-agent` entry.

Worth stating explicitly, because it is the reason the aggregate job exists: branch protection must **not** require the per-suite `Check tests (<suite>)` contexts. That job is a matrix whose membership is selected dynamically by `.github/scripts/select-checks.sh`, so those context names are not stable, and requiring them would hang any PR that legitimately skips a suite. `ci-success` gives protection exactly one fixed name to gate on.

Sequencing: STAC-25522 must land **after** this PR, otherwise protection would require a check that master does not yet produce — recreating the very deadlock it is meant to fix.

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,384 @@
...
+        run: |
+          set -eo pipefail
+          git config --global --add safe.directory '*'
+          source .setup-scripts/setup_env.sh
+
+      - name: Configure the private PyPI index (GitLab Package Registry)
+        env:
+          GITLAB_PACKAGE_REGISTRY_PYPI_SIMPLE_URL: ${{ vars.GITLAB_PACKAGE_REGISTRY_PYPI_SIMPLE_URL }}
+          GITLAB_PACKAGE_REGISTRY_USER: ${{ secrets.GITLAB_PACKAGE_REGISTRY_USER }}
+          GITLAB_PACKAGE_REGISTRY_READONLY_PASSWORD: ${{ secrets.GITLAB_PACKAGE_REGISTRY_READONLY_PASSWORD }}
```
LouisLotter, line 386: [P1] These Actions secrets are still available to PR-controlled code. The setup script writes them to ~/.netrc, which persists in this job and is readable by the subsequent checksdev test and benchmark steps; limiting the env values to this step does not limit access to the credentials. Dependabot also receives none of these Actions secrets, so any PR selecting vsphere, or a shared path that selects the full matrix, will still fail. Please remove private-index credentials from automatic PR jobs: expose the dependency through a credential-free read path, or split the private-index suite into a protected GitHub Environment with required approval after the code is reviewed.

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,384 @@
...
+      - check-tests
+      - workflow-security
+    runs-on: docker-public
+    timeout-minutes: 5
+    steps:
+      - name: Evaluate upstream job results
+        env:
+          # `skipped` is a legitimate outcome and must not be conflated with
+          # success: check-tests skips when no suite was selected, and the whole
+          # pipeline skips for fork PRs. Only `failure` and `cancelled` fail here.
```
LouisLotter: [P1] This turns a fully skipped pipeline into a successful required check. On a fork PR, select-checks, validate, check-tests, and workflow-security are all skipped, yet CI success exits zero, so the PR can satisfy branch protection without running any CI. Require select-checks, validate, and workflow-security to succeed; allow check-tests to be skipped only when a successful selector produced an empty matrix. Fork PRs should run credential-free tests on GitHub-hosted runners or remain blocked, never report success without tests. Also run this aggregate job on GitHub-hosted infrastructure: the workflow itself is PR-authored, so checking out nothing does not prevent a fork from editing the job to execute code on docker-public.

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,384 @@
...
+  # That image exists mainly to carry Docker CE and docker-compose, which only the
+  # phase-2 Splunk / checks_dev suites use; the phase-1 suites need a Python
+  # toolchain and nothing else. Dropping it also drops the registry credentials
+  # these jobs used to need -- see CREDENTIALS below.
+  #
+  # BCI publishes 3.13.13, one patch behind the 3.13.14 the agent embeds and the
+  # GitLab image pinned. CPython patch releases are bugfix-only, and the full
+  # phase-1 matrix (including vsphere against the private index) was verified
+  # green on 3.13.13 before this switch. Worth realigning when BCI ships .14.
+  BCI_PYTHON_IMAGE: registry.suse.com/bci/python:3.13@sha256:7d36dd3ba6596fb690e31d956952059fd010604ad6309f06462c02c4c9c01461 # 3.13.13
```
LouisLotter, line 163: [P2] This is now the correct BCI image, but it is pulled directly from registry.suse.com while every consumer job runs on the self-hosted docker-public runner. Self-hosted jobs must pull upstream images through registry.tooling.stackstate.io/suse/...; direct upstream pulls cross our NAT and violate the standing proxy policy. The clean fix for phase one is to run these credential-free jobs on GitHub-hosted runners and keep the direct BCI reference. If self-hosted runners remain, use the SUSE Private Registry proxy without re-exposing pull credentials to PR-controlled workflows.

LouisLotter (changes requested):
Three follow-ups remain after the first review: private-PyPI credentials are still exposed to PR-controlled code, the aggregate gate succeeds when all substantive CI is skipped, and the direct BCI pull violates the self-hosted registry-proxy policy.

LouisParkin:
Thanks @LouisLotter — both P1s were real. I've pushed 863b1cd which restructures the credential handling rather than patching around it. Details per finding:

### P1 — credentials reachable by PR-controlled code ✅ fixed

You were right that the fork guard wasn't the boundary I implied. `setup_artifact_registry.sh` writes `~/.netrc`, so scoping `env:` to that one step was cosmetic — every later step in the job, including the PR-authored test code we're about to execute, could read the file. A same-repo PR (the common case for us) had full access.

Fixed by making the split structural instead of procedural:

- `select-checks.sh` now emits **two** outputs — `checks` (14 suites, no credentials) and `private_checks` (`vsphere`, the only suite needing the private index).
- Those feed **two separate jobs**. `check-tests` no longer references any secret at all — the credential step is gone from it entirely.
- `check-tests-private-index` carries `environment: private-package-index`, a protected Environment with `prevent_self_review: true` and `dev-team` as required reviewer.

So the credential is no longer reachable from a job that runs untrusted code without a human first approving that specific run. Verified live on this PR's run — `current_user_can_approve=false` for me as the author, and the job is sitting in `waiting`.

I also made the registry setup **fail loudly** (`exit 1`) instead of warn-and-continue. Previously a missing secret produced a silent pass.

**Residual, and I want to be honest about it rather than overclaim:** the secrets are still repo-level (`GITLAB_PACKAGE_REGISTRY_USER`) and org-level (`GITLAB_PACKAGE_REGISTRY_READONLY_PASSWORD`). A future workflow edit could reference them from an ungated job. Making them environment-scoped requires an org-level change that also affects `stackstate-agent`, so I've left it — but any such edit is diff-visible in this file.

### P1 — aggregate gate green when everything skips ✅ fixed

Confirmed and worse than described. The loop only failed on `failure`/`cancelled`, so `skipped` counted as success. Combined with the fork guards that were on *every* job, a fork PR could satisfy branch protection having run no CI whatsoever.

`ci-success` now distinguishes the two meanings of "skipped":

- `select-checks`, `validate`, `workflow-security` must be `success`. No skip is legitimate.
- `check-tests` may skip **only** when the selector legitimately chose zero suites.
- `check-tests-private-index` may additionally skip when the PR is from a fork.

Anything else fails the gate.

Related: I found this repo's branch protection currently requires the `ci/gitlab/gitlab.com` context, which can no longer ever report since the GitLab project was archived — `master` is presently unmergeable, with an empty check rollup that reads as "no checks" rather than an error. Filed as STAC-25522, to land *after* this PR so `CI success` exists as a target first. That makes this gate load-bearing, which is partly why I wanted the skip semantics correct.

### P2 — BCI pulled directly rather than through the proxy 🔄 moot

This one I'd have pushed back on, but it's moot now: **every job moved to `ubuntu-latest`.** Once vsphere was the only suite needing credentials, there was no reason to keep 14 suites on self-hosted runners.

For the record on the original point — the merged precedent in `stackstate-agent` (`build-binaries.yml:301`, `build-deb.yml:380`) pulls BCI direct from `registry.suse.com` on self-hosted runners, with the comment *"the BCI stages come from registry.suse.com and need no auth"*; the proxy is used for the auth-requiring Docker Hub/Ubuntu stages. `registry.tooling.stackstate.io` appears in no GitHub Actions workflow in the org today. Happy to be corrected if that's drifted from current intent.

Also worth noting on the Dependabot sub-point: it isn't enabled on this repo, so there's nothing consuming that config today.

### Runner move — it's faster

Somewhat to my surprise, GitHub-hosted beat the self-hosted lane on this workload, presumably because we weren't queuing behind other jobs:

| | self-hosted (prev run) | `ubuntu-latest` (this run) |
|---|---|---|
| slowest suite | 172s | **99s** |
| fastest suite | 99s | **68s** |

All 16 credential-free jobs green, no changes needed to the BCI container or `zypper` setup.

Validated with actionlint, shellcheck and Zizmor (`No findings`). `select-checks.sh` unit-tested across full-run / empty-selection / single-public / vsphere-only under `set -u`.

---

🙏 **One ask:** this PR's `vsphere` job is blocked awaiting a `dev-team` approval, and by design I can't approve my own run. If someone could approve the pending deployment, that'll confirm the credentialed path still works end-to-end before merge.

LouisLotter (dismissed):

LouisParkin:
Follow-up commit b1715b4 — comments only, no behaviour change, but it corrects a factual claim reviewers would otherwise rely on.

The header said *"releases keep running from GitLab, so the GitLab pipeline stays live."* That is no longer true. The GitLab project (`stackvista/agent/stackstate-agent-integrations`, id 13169244) reports `archived: true`, last activity 2026-07-20. Archived projects are read-only, so `publish-checks-dev` cannot run there either.

Two consequences worth being explicit about, because they change the deferred items from "scheduled" to "currently broken":

- the five unported suites (`splunk_{base,health,metric,topology}`, `stackstate_checks_dev`) have **no CI on any platform** right now;
- there is **no way to publish a new `stackstate_checks_dev`** until publishing has a home.

Neither is caused by this PR — this PR is what restores coverage for the other 15 — but the header shouldn't imply a safety net that isn't there.

I also dropped a second stale claim: phase 2 no longer rides on an ARC DinD sidecar, since every job now runs on `ubuntu-latest`, which already provides a Docker daemon. Phase 2 just needs a docker client in the job image.

Deferred work now carries ticket references rather than vague pointers:

| | |
|---|---|
| STAC-25531 | the five Docker-daemon suites |
| STAC-25532 | `publish-checks-dev` |
| STAC-25533 | Cerberus failure notification |

actionlint, shellcheck and Zizmor clean; selector smoke-tested (14 + 1 unchanged).

LouisParkin:
@LouisLotter apologies — you approved this at 08:19 and I knocked it straight back out at 08:37.

This base branch has `dismiss_stale_reviews: true` (and `require_last_push_approval: true`), so my follow-up push auto-dismissed your approval. That push was **comments only** — `b1715b4`, +23/−16 across the two CI files, no behaviour change. It corrects a claim in the header that I'd discovered was false: the header said *"releases keep running from GitLab, so the GitLab pipeline stays live"*, but the GitLab project is archived (`archived: true`, last activity 2026-07-20), so the deferred jobs can't run there either. I also swapped the vague "phase 2 / a follow-up" pointers for STAC-25531/25532/25533.

Rebuilding your approval on a comment-only diff is a poor trade for your time, and I should have batched it with the review response instead of pushing separately. Re-requesting now.

Nothing else changed since you approved: still 16/16 credential-free jobs green, and the vsphere job is still gated behind the `private-package-index` environment.

One thing that still needs a **second** person regardless of the approval: the run is sitting in `waiting` on a `dev-team` approval for the vsphere deployment, and `prevent_self_review` means I can't release it myself. Whoever re-approves the PR can also approve that pending deployment on run [30989773234](https://github.com/StackVista/stackstate-agent-integrations/actions/runs/30989773234) — that's the last unproven path before merge.

LouisParkin:
### `private-package-index`: disabled `prevent_self_review`, and a better fix for the P1

@LouisLotter — heads-up on a change to the control I added for your P1 about credentials reaching PR-authored code, plus a proposal that I think closes it properly.

**What changed:** `prevent_self_review` on the `private-package-index` environment is now `false`. Required reviewers (`dev-team`) and the job's fork guard are unchanged.

**Why.** The gate as configured made the credentialed `vsphere` leg a *required* check that the PR author structurally could not satisfy. Because `SHARED_PATHS` includes the CI files, every CI-touching PR selects the full matrix including `vsphere` — roughly 16% of commits on this repo over the last year — so the author needed a second person on each one. That was my configuration choice, not something your review asked for.

The threat it was carrying is already handled twice for the case that actually matters:

- the job's `if:` requires `head.repo.full_name == github.repository`, so fork PRs skip it; and
- GitHub does not expose Actions secrets to fork `pull_request` runs on public repos at all.

So the environment gate only ever constrained the 27 accounts that already hold push access — and write access to a repo is already trust over its secrets under GitHub's model. Your narrower point stands (an in-repo branch is unreviewed at the time CI runs), but a self-approvable gate answers that with an audit record rather than an access boundary, which is what it was really providing.

**The proper fix, which I'll do next.** Your first suggested remedy — a credential-free read path — turns out to be the right one, just not by re-hosting. Context I got wrong earlier: `vsphere-automation-sdk==1.82.0` is not something we patch. It is an unmodified upstream wheel that VMware withdrew from public PyPI, self-hosted in our GitLab registry only because that org was private. Upstream carries no SPDX license and the version predates their current tags, so publishing it publicly is a licensing question, not an engineering one — I'm not proposing that.

Instead I'll split the credential away from PR code:

1. a `fetch-private-deps` job that checks out no executable repo content and uses a **hardcoded** package spec, so nothing PR-authored can influence it, pulls the wheel from the GitLab registry, and uploads it with `actions/upload-artifact`;
2. the `vsphere` job downloads that artifact and installs `--no-index` offline.

The credential then never shares a job with PR-authored code, nothing is redistributed, the environment gate can be deleted, and `vsphere` rejoins the normal matrix — including on fork PRs, which currently skip it and lose that coverage.

I'll raise that as a separate PR rather than pushing here, since `dismiss_stale_reviews` + `require_last_push_approval` mean any further push to this branch re-dismisses your review.

LouisParkin:
### Status: fully green, and the environment gate is gone

@LouisLotter — #364 merged into this branch, so what you'll be reviewing has changed materially since your last pass. Summary of where your P1 now stands.

**Your finding:** `setup_artifact_registry.sh` wrote the registry password to `~/.netrc`, which persisted for the rest of the job and was readable by the subsequent `checksdev test` and benchmark steps — i.e. by PR-authored code. You offered two remedies: a credential-free read path, or a protected environment with required approval.

**I originally took the second.** It worked but was the wrong trade: `SHARED_PATHS` covers the CI files, so it fired on ~28 of the last 176 commits (~1 in 6) and structurally blocked authors on their own PRs.

**This branch now implements your first remedy instead.** `.setup-scripts/fetch_private_wheels.sh` writes the netrc, downloads a **hardcoded** package set, deletes the netrc, and repoints `~/.pip/pip.conf` at a local wheelhouse via `find-links`. The suite then installs offline with no credential on disk and no authenticated index configured. `environment: private-package-index` is removed.

Four details are load-bearing, and I'd welcome your eye on them specifically:
- the package set is hardcoded rather than read from the tree, so a PR cannot redirect pip while the credential is on disk;
- `--only-binary=:all:`, because downloading an sdist executes its `setup.py`;
- a **system** interpreter, resolved from candidates and then **rejected if it resolves inside `GITHUB_WORKSPACE`** — the toolchain venv is built by repo code, so its pip must never be in the credential's path;
- an `EXIT` trap revokes the credential on failure paths, then the script asserts the netrc is gone and the wheel actually arrived rather than silently falling through to the PyPI `0.0.1` placeholder.

**Latest run [31009548529](https://github.com/StackVista/stackstate-agent-integrations/actions/runs/31009548529): all 19 jobs green, `vsphere` included, zero pending deployment approvals.**

**Two things I want to be straight about:**
- **Fork PRs still skip `vsphere`.** I claimed earlier they'd regain coverage; that was wrong. Forks get no Actions secrets on a public repo, so they can't fetch the wheel at all. The fork guard stays.
- **Residual risk is unchanged and stated in the workflow:** a PR that edits the workflow itself can still reach the secret, because `pull_request` runs the PR's own copy. That's true of every secret in every repo and is contained by review, not CI. What changed is that ordinary test code and its dependency tree no longer see it — and that needed no workflow edit at all.

Also, apologies again for dismissing your approval this morning; that was my push, not you.

Ready for another look whenever suits.

.github/workflows/checks-tests.yml
```diff
@@ -0,0 +1,537 @@
...
+        #
+        # NOTE: pip.conf is read from $HOME, so tox must still pass HOME into the
+        # testenv. tox drops every variable absent from `passenv`, and pip then
+        # resolves `~` from the passwd database rather than the environment --
+        # which points at the wrong home in a container job, where HOME is
+        # /github/home. Without it the suite silently falls back to public PyPI
+        # and installs the 0.0.1 placeholder. See vsphere/tox.ini.
+        run: |
+          set -eo pipefail
+          .setup-scripts/fetch_private_wheels.sh "${RUNNER_TEMP}/private-wheels"
```
LouisLotter, line 408: [P1] This still executes PR-controlled code with registry secrets in scope. The job checks out the PR merge commit and sources its setup code, then invokes this helper from the same checkout with all three credentials in the environment. A same-repository PR can modify the helper—or earlier setup to poison the runtime—and exfiltrate them before the netrc cleanup. Deleting the netrc only protects subsequent test steps. Please remove secrets from pull_request execution, or put environment-scoped credentials behind non-author approval; repo/org secrets plus a PR-controlled helper do not create a security boundary.

LouisLotter (approved):

.github/workflows/checks-tests.yml
LouisParkin: You're right, and this wasn't paranoia — thanks for pushing on it a second time.

Worse, the comment you were replying to was actively misleading. It claimed reaching the secret "would require editing this workflow". It doesn't. Two cheaper paths, both of which you identified:

1. `fetch_private_wheels.sh` is itself checked out from the PR head, so the PR can edit the very script that holds the credential.
2. `source .setup-scripts/setup_env.sh` runs **as root, in the container, before** the credential step — exactly the "poison the runtime" case you name. It can overwrite `/usr/bin/python3.13` outright, which defeats the workspace-interpreter guard I'd written to make "system interpreter" a guarantee. That guard only ever proved the binary sits outside the checkout, not that it's the one SUSE shipped.

And the general form of your last sentence is the part I should have started from: a `pull_request` run executes the PR's own copy of the workflow *and* every script it calls, so no arrangement of code inside the run can protect a secret the run is holding. Confining the credential to one step is hardening, not a boundary. I'd labelled it as a boundary. That's corrected.

**Fix: #365**, taking your first option. The private-index suites no longer run on `pull_request` at all — they run on `push`/tag/`workflow_dispatch`, where the code has been reviewed before it can see the secret. `select-checks.sh` moves them to a `deferred_private_checks` output so the matrix is empty rather than gated; `fetch_private_wheels.sh` refuses outright on a `pull_request` event, so the rule survives an edit to the workflow condition instead of depending on it; and `ci-success` reports the deferral so a green PR never implies vsphere was covered.

I went with option 1 over your option 2 (environment-scoped credentials behind non-author approval) because the environment gate is what we removed earlier in this PR: `SHARED_PATHS` covers the CI files, so it fired on roughly one commit in six, and it only ever constrained people who already have write access. Option 1 removes the exposure outright rather than gating it.

The honest cost is that vsphere is verified on the release branch rather than on the PR that changes it. `vsphere/` changes a handful of times a year and most of its current runs are triggered by unrelated CI edits, so I think that's the right trade for now — but it is a real gap, not a nil one.

One thing that came out of digging into this, which makes the whole argument temporary: **the private index looks avoidable entirely.** VMware now publishes this SDK to public PyPI under renamed packages — `vmware-vapi-runtime`, `vmware-vapi-common-client`, `pyvmomi`, all at 9.1.0.0 — and serves the NSX/VMC wheels from its own public PEP 503 index. The names we're blocked on (`vsphere-automation-sdk`, `vapi-runtime`) are squatted `0.0.1` placeholders. We're on the private index purely because of the ancient `vsphere-automation-sdk==1.82.0` pin. Modernising it deletes the credential, this job and the coverage gap in one change. Ticket to follow; I didn't want to fold an SDK upgrade into the CI migration.

I'd rather you re-read #365 than take my word for it, given I got this wrong once already.

LouisLotter (dismissed):

LouisParkin:
@LouisLotter — agreed on the call, and #365 is merged into this branch. Thread resolved.

## Your one remaining condition, checked

> The only reason to continue blocking would be if the GitLab token has permissions beyond read-only access to project `71271774`.

Verified against the GitLab API rather than taken on trust. Project `71271774` is `stackvista/ci/packages` (private). Its deploy tokens:

| token | active | scopes |
|---|---|---|
| `CI Package registry access` | **revoked** | `read_package_registry`, `write_package_registry` |
| `gitlab-proxy` | **revoked** | `read_package_registry` |
| `gitlab-proxy-read-package-registry` (13698947) | **revoked** | `read_package_registry` |
| `gitlab-proxy-read-package-registry` (13702516) | **revoked** | `read_repository`, `read_package_registry` |
| `gitlab-proxy-read-package-registry` (13702666) | **active** | **`read_package_registry`** |

**The only live token carries `read_package_registry` and nothing else** — no `write_package_registry`, no `read_repository`. The one token that ever had write scope is revoked. So your condition is not met, and there's no reason to keep blocking. (Caveat for completeness: I can't read the secret's value, so I'm matching on the project's token inventory rather than proving which credential the secret holds. If it were a group/personal token instead of a deploy token, this wouldn't cover it.)

One thing worth recording that *supports* moving on: `GITLAB_PACKAGE_REGISTRY_READONLY_PASSWORD` is an **org-level secret with `visibility: all`** ([`orgVariables/resources_github.yaml:47`](https://github.com/StackVista/pulumi-infra/blob/main/github/orgVariables/resources_github.yaml#L47)), so every repo in the org already receives it. Gating it in this one repo could never have contained it — which is a further argument that the environment gate was buying less than it cost. This repo receives only the simple URL and the username; the write pair (`GITLAB_PACKAGE_REGISTRY_PASSWORD`) is not granted here.

Follow-up ticket filed as promised: **STAC-25544** — move vsphere to the publicly-published VMware SDK packages (`vmware-vapi-runtime`, `vmware-vapi-common-client`, `pyvmomi`, all 9.1.0.0 on public PyPI, plus VMware's own public PEP 503 index for the NSX/VMC wheels). That deletes the credential, the `check-tests-private-index` job, and the PR coverage gap in one change. We're only on the private index because of the ancient `vsphere-automation-sdk==1.82.0` pin.

## Sorry — one more approval needed

Merging #365 pushed to this branch, so `dismiss_stale_reviews` + `require_last_push_approval` dismissed your approval for the third time. CI is re-running now. Nothing has changed since you approved except the #365 contents you already reviewed.

Worth fixing as part of **STAC-25522** (branch protection): `required_status_checks.contexts` is currently empty on `stackstate-7.78.2`, so `CI success` isn't actually enforced as a required check yet — which is the thing that would let this settle down.

LouisLotter (approved):

PR #366, opened 2026-08-06
STAC-25533: report release-branch CI failures to Slack via Cerberus

LouisParkin:
This pipeline has never had a failure notification — the retired `.gitlab-ci.yml` had no notify job and no `.cerberus` directory, so a broken release branch is silent. STAC-25510 is what that costs: process-agent's image publishing broke and nobody noticed for 12 days.

Adds a reusable `cerberus-notify.yml` (structurally identical to the one process-agent gained in STAC-25519) and calls it as a terminal job off `ci-success`.

Worth a look:

- **Push events only.** Given the existing `push:` trigger that means the release branch and release tags. PR failures already have an owner watching, and on a public repo anyone can open a deliberately-failing PR — notifying on those hands the CI channel a spam vector. Keyed on `github.event_name` rather than a hardcoded ref so the next release-branch bump touches one place, not two.
- **`action: notify`, never `block`.** Blocking locks the branch, needs the Cerberus App installed here, and mutates the branch protection pulumi-infra now owns (STAC-25522).

Depends on StackVista/pulumi-infra#277 for the repo-level secrets (the org copies are `visibility=private`; this repo is public). Until that merges the job warns and exits 0 rather than adding a second red job to an already-failed run.

Validated: actionlint + shellcheck and Zizmor 1.28.0 clean; payload dry-run produces valid JSON for both a normal commit and the empty-`head_commit` tag case.

https://stackstate.atlassian.net/browse/STAC-25533


.github/workflows/cerberus-notify.yml
```diff
@@ -0,0 +1,122 @@
...
+          # Empty on tag pushes, which carry no head_commit. COMMIT_TITLE below
+          # falls back to the sha so the Slack message is never blank.
+          COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
+          SUITE: ${{ inputs.suite }}
+        run: |
+          set -euo pipefail
+
+          if [ -z "${CERBERUS_LAMBDA_URL}" ] || [ -z "${CERBERUS_API_TOKEN}" ]; then
+            echo "::warning title=Cerberus not configured::CERBERUS_LAMBDA_URL and/or CERBERUS_API_TOKEN is not visible to this repo, so the ${SUITE} failure was not reported to Slack. Both are provisioned as repo-level secrets by pulumi-infra (STAC-25533)."
+            exit 0
```
LouisLotter, line 86: [P2] Fail closed when Cerberus is unconfigured

This reports success precisely when no Slack alert was sent. The warning is only visible inside the unattended failed run, so missing or drifted secrets silently disable the safeguard. Please merge and apply StackVista/pulumi-infra#277 first, then make these secrets required or return non-zero here.

LouisLotter (approved):

.github/workflows/cerberus-notify.yml
LouisParkin: Agreed, and fixed in #369 (STAC-25555) rather than here, so this PR can merge.

I took the "return non-zero" arm rather than the "make these secrets required" arm. `required: true` is worse than it looks: a caller passing `${{ secrets.X }}` for a secret the repo does not hold yields an empty string, and GitHub rejects that as "secret not provided" — failing the *call* before any step runs, with an opaque error and no annotation naming what to fix. Keeping `required: false` and checking in the step fails just as closed, and says which secret is missing and who provisions it.

Per your sequencing point, #369 is a **draft** and stays held until StackVista/pulumi-infra#277 is merged *and applied* — that is what grants the two secrets. Merging it before then would put a second red job on every failed release-branch run that nobody could fix from this repo.

PR #367, opened 2026-08-06
STAC-25531: run the five Docker-daemon test suites in GitHub Actions

LouisParkin:
Phase 2 of STAC-25142. The four `splunk_*` suites and `stackstate_checks_dev` need a live Docker daemon, so phase 1 left them out — they currently run **nowhere**.

**Why a separate job, not the existing matrix.** `get_docker_hostname()` returns `localhost` unless `DOCKER_HOST` is set, and compose publishes ports on the Docker *host*. The test process therefore has to share the daemon's network namespace, which rules out the BCI job container. GitLab hid this behind a `docker:dind` service whose hostname resolved the same from both sides. So: `check-tests-docker` runs directly on the runner with `actions/setup-python`.

**Registry change.** All five fixtures pulled through `registry.tooling.stackstate.io`, which answers `401` publicly — a credential this public repo's `pull_request` jobs must not hold. Every path was a `/docker/` proxy-cache entry, so they now point at the Docker Hub originals via `${SPLUNK_IMAGE}` / `${VAULT_IMAGE}`, defaulted publicly and overridable back to the proxy.

`splunk_base` is a build dependency of the other three splunk suites, so the selector fans a change to it out to all four.

Stacked on #366 — retarget to `stackstate-7.78.2` once that merges.

Validated: actionlint + shellcheck + Zizmor clean; selector routing exercised in a scratch worktree for push, `splunk_base`, `splunk_metric`, and `kubelet` changes.


LouisParkin:
Green — all five suites pass on [run 31108281630](https://github.com/StackVista/stackstate-agent-integrations/actions/runs/31108281630). Splunk comes up well inside the timeout on a hosted runner, and runner Python 3.13 behaves the same as the BCI 3.13.13 the matrix uses.

Getting there took two fixes the first run exposed, both in the toolkit rather than the workflow:

1. **Compose v1 is gone.** All five failed with `FileNotFoundError: 'docker-compose'`. The standalone v1 binary reached EOL in July 2023 and is absent from current runner images, which ship v2 as `docker compose`. The GitLab runner image still carried v1, which is why this never surfaced there. `compose_command()` now resolves it, preferring v2 and falling back to v1 so anyone still on an older toolchain keeps working; `DOCKER_COMPOSE_COMMAND` overrides.

2. **`compose_file_active()` parsed a v1-only artefact** — it scanned the `ps` table for the row of dashes v1 printed under the header. v2 does not emit one, so it returned `False` unconditionally and two `stackstate_checks_dev` tests still failed. It now uses `ps -q`, which is stable across both generations. Verified against a live daemon: `False` before up, `True` after up, `False` after down.

PR #368, opened 2026-08-06
STAC-25544: move vsphere to the publicly-published VMware SDK

LouisParkin:
vsphere pinned `vsphere-automation-sdk==1.82.0`, which VMware never published to public PyPI — the name is squatted there by an unrelated `0.0.1` placeholder — so it was mirrored into the private GitLab package registry and pulled with a credential.

That credential could not be held safely on a public repo. A `pull_request` run executes the PR's own copy of the workflow and every script it calls, so a run holding a secret can always be made to disclose it. The only sound answer was to withhold it, which meant **vsphere never ran on pull requests** and was verified only after merge.

VMware publishes the same SDK to public PyPI under renamed packages, actively maintained (all `9.1.0.0`, May 2026): `pyvmomi`, `vmware-vcenter`, `vmware-vapi-runtime`, `vmware-vapi-common-client`. Between them they satisfy every import the check makes — `pyVmomi`/`pyVim`, `vmware.vapi.*` and `com.vmware.*`.

So the pin, the credential, the separate job and the coverage gap all go together. `fetch_private_wheels.sh`, `PRIVATE_INDEX_CHECKS`, and `check-tests-private-index` are deleted; vsphere joins the ordinary matrix and runs on PRs like everything else.

**No job in this workflow now has a secret in scope** — a much simpler property to maintain than any arrangement that held one.

Verified: the four pins resolve from public PyPI alone, and the full vsphere suite passes against them (58 passed). actionlint, shellcheck, Zizmor clean.

Stacked on #367. Once this merges, the `GITLAB_PACKAGE_REGISTRY_*` grants can come off this repo in pulumi-infra.


LouisParkin:
Green on [run 31111900999](https://github.com/StackVista/stackstate-agent-integrations/actions/runs/31111900999).

The one that matters: **`Check tests (vsphere)` ran and passed on a pull request**, in the ordinary matrix, with no credential anywhere in the run. That is the coverage gap closing — until now vsphere was only ever verified after merge.

One fix on the way: `checksdev validate dep` cross-checks each suite's pins against `agent_requirements.in`, the manifest of what the agent embeds, which still named the old SDK. The four public packages replace that single entry there too.

